PR #39, opened 2022-09-04
Feature/api v1 settings member attributes

epipav:
# Changes proposed ✍️

- member attribute settings repository, service and api layers
- predefined attributes from integrations

        
## Checklist ✅
- [x] Tests are passing.  
- [x] Environment variables have been updated
  - [x] Front-end: `frontend/.env.dist`
  - [x] Backend: `backend/.env.dist`, `backend/.env.dist.staging`, `backend/.env.dist.staging`.
  - [x] [Configuration docs](https://docs.crowd.dev/docs/configuration) have been updated.
  - [x] Team members only: update environment variables in Password manager and update the team


backend/src/database/attributes/member/enums.ts
```diff
@@ -0,0 +1,26 @@
+export const MemberAttributes = {
```
themarolt, line 1: Why is this file named `enums.ts` if it doesn't contain an enum?
Perhaps do something like it's done in `platforms.ts` and `platformDisplayNames.ts` files.

backend/src/database/repositories/memberAttributeSettingsRepository.ts
```diff
@@ -0,0 +1,230 @@
...
+import Error404 from '../../errors/Error404'
+import SequelizeFilterUtils from '../utils/sequelizeFilterUtils'
+
+const Op = Sequelize.Op
+
+class MemberAttributeSettingsRepository {
+  static async findById(id, options: IRepositoryOptions) {
+    const transaction = SequelizeRepository.getTransaction(options)
+
+    const include = []
```
themarolt, line 14: I would remove this `include` property since it's not required by `findOne` method in order to make the method simpler and not confuse other coders when they encounter it.

backend/src/database/repositories/memberAttributeSettingsRepository.ts
```diff
@@ -0,0 +1,230 @@
...
+  static async create(data, options: IRepositoryOptions) {
+    const currentUser = SequelizeRepository.getCurrentUser(options)
+
+    const tenant = SequelizeRepository.getCurrentTenant(options)
+
+    const transaction = SequelizeRepository.getTransaction(options)
+
+    const record = await options.database.memberAttributeSettings.create(
+      {
+        ...lodash.pick(data, ['type', 'name', 'label', 'canDelete', 'show']),
```
themarolt, line 52: This is not very typescripty :) I would suggest introducing an interface for `data` parameter and instead of doing `...lodash.pick` I would just set the property one by one - that way if we change the interface in the future the build will fail here if the property of the interface is removed for instance.

[Here](https://github.com/CrowdDotDev/crowd.dev/pull/43/files#diff-965f8e04f52743e99d3dcf965807f68f2d41ae8d7e726830eb77a69a4d45a524R22) is my implementation of `automationRepository` for reference.

backend/src/database/repositories/memberAttributeSettingsRepository.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    if (record.canDelete) {
+      await record.destroy({
+        transaction,
+      })
+    }
+  }
+
+  static async findAndCountAll(
+    { filter, limit = 0, offset = 0, orderBy = '' },
```
themarolt, line 124: Same here - introduce a typescript interface so that whoever calls this function knows what is expected from `filter` parameter.

backend/src/serverless/integrations/workers/devtoWorker.ts
```diff
@@ -30,6 +35,20 @@ async function devtoWorker(body: DevtoIntegrationMessage) {
     // Inject user and tenant to IRepositoryOptions
     const userContext = await getUserContext(tenant)
 
+    const memberAttributesService = new MemberAttributeSettingsService(userContext)
+
+    await memberAttributesService.createPredefined(DevtoMemberAttributes)
+
+    await memberAttributesService.createPredefined(
+      TwitterMemberAttributes.filter((a) => a.name === MemberAttributes.URL.name),
```
themarolt, line 43: Maybe create a helper method to just do `pickAttributes(names, arrayOfAttributes)` and call it like this (where `MemberAttributes` is an actual enum as well): `pickAttributes([MemberAttributes.URL], TwitterMemberAttributes)`.

backend/src/services/memberAttributeSettingsService.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  async createPredefined(attributes: Attribute[]) {
+    const transaction = await SequelizeRepository.createTransaction(this.options.database)
+
+    try {
+      const created = []
+      for (const attribute of attributes) {
+        // check attribute already exists
+        const existing = await MemberAttributeSettingsRepository.findAndCountAll(
```
themarolt, line 49: What happens when we remove an attribute from predefined ones?

backend/src/serverless/integrations/workers/twitterWorker.ts
```diff
@@ -78,6 +80,8 @@ async function twitterWorker(body: TwitterIntegrationMessage): Promise<TwitterOu
 
     const followersSet: Set<string> = new Set(integrationUpdated.settings.followers)
 
+    await new MemberAttributeSettingsService(userContext).createPredefined(TwitterMemberAttributes)
```
themarolt, line 83: If I understand this correctly - every time a worker is being run we again try to create attribute settings in the database?
Doesn't seem like the best way to do it - perhaps it would be better if we do this just once when integration is created or something like that and then prepare a way to do it on demand to run it when we manually change the predefined attributes? It seems like a lot of processing for something that will be actually done in less than 1% of the runs.

backend/src/services/memberAttributeSettingsService.ts
```diff
@@ -0,0 +1,161 @@
...
+import { IServiceOptions } from './IServiceOptions'
+
+export default class MemberAttributeSettingsService {
+  options: IServiceOptions
+
+  constructor(options) {
+    this.options = options
+  }
+
+  async create(data) {
```
themarolt, line 15: Again an interface would be nice.

backend/src/services/memberAttributeSettingsService.ts
```diff
@@ -0,0 +1,161 @@
...
+        error,
+        this.options.language,
+        'memberAttributeSettings',
+      )
+
+      throw error
+    }
+  }
+
+  async destroyAll(ids) {
```
themarolt, line 81: We should use some typing information like `ids: string[]`.

backend/src/services/memberAttributeSettingsService.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      await SequelizeRepository.commitTransaction(transaction)
+    } catch (error) {
+      await SequelizeRepository.rollbackTransaction(transaction)
+      throw error
+    }
+  }
+
+  async update(id, data) {
```
themarolt, line 99: interfaces and types :)

backend/src/api/settings/memberAttributeCreate.ts
```diff
@@ -0,0 +1,16 @@
+import PermissionChecker from '../../services/user/permissionChecker'
```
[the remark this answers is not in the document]
epipav: openapi comments included

backend/src/database/repositories/memberAttributeSettingsRepository.ts
epipav: redundant includes removed

backend/src/serverless/integrations/workers/devtoWorker.ts
epipav: it'll be very uncommon case (most probably never happened before) so I'm not sure how to handle it, because there'll most likely be data already populated in member.attributes with these fields

epipav: helper function included

backend/src/services/memberAttributeSettingsService.ts
epipav: it'll be very uncommon case (most probably never happened before) so I'm not sure how to handle it, because there'll most likely be data already populated in member.attributes with these fields

backend/src/serverless/integrations/workers/twitterWorker.ts
epipav: Now we're checking integration.settings.updateMemberAttributes boolean to decide update or not, after this function updateMemberAttributes is set to false again

epipav:
@themarolt We're setting integration.settings.updateMemberAttributes on integration connect now :+1: Merging this branch

PR #56, opened 2022-09-15
Feature/advanced filtering

joanreyero:
# Changes proposed ✍️
- Added advanced filtering
- Replaced all relevant repositories with advanced filtering
- Normal filtering still works
        
## Checklist ✅
- [ ] Label appropriately with `type:feature 🚀`, `type:enhancement ✨`, `type:bug 🐞`, or `type:documentation 📜`.
- [x] Tests are passing.  


backend/src/services/aws.ts
```diff
@@ -63,12 +63,13 @@ export async function detectSentiment(text) {
...
     return {
-      sentiment: fromAWS.Sentiment.toLowerCase(),
+      label: fromAWS.Sentiment.toLowerCase(),
       positive: fromAWS.SentimentScore.Positive,
       negative: fromAWS.SentimentScore.Negative,
       neutral: fromAWS.SentimentScore.Neutral,
       mixed: fromAWS.SentimentScore.Mixed,
-      score: fromAWS.SentimentScore.Positive - fromAWS.SentimentScore.Negative,
+      // Mapping the value from -1,1 to 0,10
+      sentiment: 5 * (fromAWS.SentimentScore.Positive - fromAWS.SentimentScore.Negative + 1),
```
epipav, line 72: I personally think sentiment.score is more clear than sentiment.sentiment here -- Did we change it so it's not mixed with member.score?

backend/src/database/repositories/filters/queryTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+  orderBy: any
+  limit?: string | number
+  offset?: string | number
+  fields?: false | string | string[]
+  include?: any
+}
+
+export interface QueryOutput {
+  ['where']?: any
+  ['having']?: any
```
epipav, line 12: Out of curiosity: what's the difference here between using it as `where?: any` vs` ['where']: any `?

epipav (commented):
Looking good :fire:  I've included comments and some nitpicks

backend/src/database/repositories/filters/queryTypes.ts
joanreyero: There does not seem to be one 😄
Moving to `where?: any`

backend/src/services/aws.ts
joanreyero: We want to use this as the default. I think it is more clear to the API user this way.

backend/src/database/repositories/reportRepository.ts
```diff
@@ -199,34 +201,49 @@ class ReportRepository {
...
 
-    const where = { [Op.and]: whereAnd }
+    const parser = new QueryParser(
+      {
+        nestedFields: {
+          sentiment: 'sentiment.sentiment',
+        },
+      },
+      options,
+    )
```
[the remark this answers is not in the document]
joanreyero: Because whenever we send a sentiment filter (without specifying which specific mood), we should be using `sentiment.sentiment` to filter. This makes the UX better, since API users only need to sent `sentiment`, not `sentiment.sentiment`. The rest is of the fields are because AWS is returning it, and it is impossible to map everything to 1 value without loosing information (3 dimensions vs 1).

PR #62, opened 2022-09-22
Feature/member attribute filters

epipav:
# Changes proposed ✍️
- member filtering now returns new fields: lastActive, avgSentiment, activityCount
- member filtering no longer returns list of activities for each member
- member filtering is now using advanced filters
- tests for dynamic attribute filtering & sorting
- more tests for dynamic member attribute settings functions

## Checklist ✅
- [ ] Label appropriately with `type:feature 🚀`, `type:enhancement ✨`, `type:bug 🐞`, or `type:documentation 📜`.
- [ ] Tests are passing.  

### DIFF
--- a/backend/src/database/repositories/memberRepository.ts
+++ b/backend/src/database/repositories/memberRepository.ts
@@ -407,7 +407,6 @@ class MemberRepository {
     },
     options: IRepositoryOptions,
   ) {
-    const literalReachTotal = Sequelize.literal(`("member".reach->'total')::int`)
 
     let customOrderBy: Array<any> = []
 
@@ -446,9 +445,11 @@ class MemberRepository {
       SequelizeFilterUtils.customOrderByIfExists('averageSentiment', orderBy),
     )
 
-    customOrderBy = customOrderBy.concat(
-      SequelizeFilterUtils.customOrderByIfExists(literalReachTotal, orderBy),
-    )
+    if (orderBy.includes('reach')) {
+      customOrderBy = customOrderBy.concat(
+        [Sequelize.literal(`("member".reach->'total')::int`), orderBy.split('_')[1]]
+      )
+    }
 
     if (!advancedFilter) {
       advancedFilter = { and: [] }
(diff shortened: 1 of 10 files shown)


backend/src/database/repositories/memberRepository.ts
```diff
@@ -394,276 +397,364 @@ class MemberRepository {
...
+            column: 'username',
+          },
+        },
+      },
+      options,
+    )
+
+    const parsed: QueryOutput = parser.parse({
+      filter: advancedFilter,
+      orderBy: orderBy || ['createdAt_DESC'],
```
joanreyero, line 706: This should be `joinedAt`

backend/src/database/repositories/memberRepository.ts
```diff
@@ -785,6 +880,24 @@ class MemberRepository {
       transaction,
     })
 
+    output.lastActive = output.activities[0]?.timestamp ?? null
```
joanreyero, line 883: I like this lastActive. However, the frontend needs also the lastActivity (timestamp, platform, type) ([Notion](https://www.notion.so/crowddev/Design-and-front-end-for-API-V1-62cc346e6677405ba33ec0e7f9b9b056#93430b2e9ea34324bae695b7a729d764)). I think we could do both.

backend/src/database/repositories/memberRepository.ts
epipav: changed to joinedAt :+1:

backend/src/database/repositories/memberRepository.ts
epipav: added whole last activity object as well.. But we're lazy loading this activity along with organization and tasks as well. We'll need to do some performance tests on it, the overhead of additional queries may slow down the response time when getting a lot of members from member.list

epipav:
- On tenant create we're adding default member attributes to the member attribute settings now.  Default member attributes has: bio, location, url and job title
- member.attributes.id renamed to member.attribtes.sourceId
- When creating member attribute settings, we're now checking if the attribute name already exists as a fixed member field. If it is, we're throwing a 400 error

PR #71, opened 2022-10-06
Frontend runtime configuration mode for self/cloud hosted image

themarolt:
# Changes proposed ✍️
- When building frontend image for self/cloud hosting we don't want baked in env. variables so the change changes nginx-alpine docker image entrypoint to first replace env. variables in the code with actual values from the environment at runtime of the image so we can have an image without secrets and pushed to a public dockerhub repository.
        
## Checklist ✅
- [ ] ~Tests are passing.~  
(https://docs.crowd.dev/docs/updating-api-documentation)).  


mariobalca (approved):
Looks good to me! Thanks for handling this 🙏 

One minor thing that I noticed is that we don't have the `VUE_APP_STRIPE_PUBLISHABLE_KEY`within the **.env**, but for now I don't think that will be a problem since we don't stripe properly configured and being used.

Hopefully we can tackle that soon enough 💪

PR #95, opened 2022-10-21
Moved Python microservices to new API

joanreyero:
# Changes proposed ✍️
- Moved Merge suggestions to new API
  - Refactored how the service works, such that instead of having a coordinator and checking merge suggestions periodically, we call the service every time a member is added to do a check against the whole database.
  - This is possible since usernames are static, we did not actually need a coordinator
  - If a username changes, we are adding a new member anyway, so we are not loosing that edge case
- Moved engagement score to new API
- Removed some unnecessary modules in Python
- Fixed a typo in `workerTypes`
        
## Checklist ✅
- [ ] Tests are passing.  


backend/config/custom-environment-variables.json
```diff
@@ -99,7 +99,7 @@
     "clientSecret": "CROWD_GOOGLE_CLIENT_SECRET",
     "callbackUrl": "CROWD_GOOGLE_CALLBACK_URL"
   },
-
```
themarolt, line 102: Maybe just remove this blank line so we don't have a random blank line in the middle =)

backend/src/bin/jobs/index.ts
```diff
@@ -16,9 +14,7 @@ const jobs: CrowdJob[] = [
   discordCoordinator,
   slackCoordinator,
   weeklyAnalyticsEmailsCoordinator,
-  checkMemberScoresCoordinator,
-  checkMembersToMergeCoordinator,
-  checkSqsQueues,
```
themarolt, line 21: `checkSqsQueues` was removed - was this intentional?

backend/src/serverless/microservices/python/crowd-members-score/crowd/members_score/worker.py
```diff
@@ -6,4 +6,4 @@ def members_score_worker(tenant_id):
 
 
 if __name__ == "__main__":
-    members_score_worker("ab1d6621-9d56-411a-92ad-cd24a2185cd7")
```
themarolt, line 9: Why is this uuid hardcoded to begin with?

themarolt (changes requested):

themarolt:
@joanreyero also tests are failing.

backend/src/serverless/microservices/python/crowd-members-score/crowd/members_score/worker.py
joanreyero: Everything under `if __name__ === "__main__"` only gets executed when calling the file directly, so this will never run. But will remove anyway 😄

backend/src/bin/jobs/index.ts
joanreyero: Nope. But I wonder how it works anyway when doing `clean-start-dev` 🤔

themarolt: Locally `checkSqsQueues` shouldn't work since slack token to post slack messages is not configured. I was asking if you intentionally removed checkSqsQueues job here otherwise please put it back in :D I think it was a merge issue.

joanreyero: Ah, ok! Yeah, putting it back

PR #197, opened 2022-11-10
Tasks module updates 

epipav:
# Changes proposed ✍️
- now tasks can have multiple assignee users. Relation is kept in `taskAssignees` table. Also removes the `assignedToId` field from tasks.
- new endpoint `POST /tenant/:tenantId/task/batch` that lets making batch operations to the task entity. Currently only supported batch action is `findAndUpdateAll`. An example payload to this operation can be:
```
{
      operation: 'findAndUpdateAll',
      payload: {
               filter: {
                      status: 'done'
               },
               update:{
                     status: 'archived'
               }
      }
} 
```
This payload will update all the tasks in a tenant that has `status:done` to `status:archived`. Filter object can be any filter that's supported by list and query endpoints. Endpoint only returns number of rows updated.
- new field `type` in tasks table. This field can have two different values: `regular` and `suggested`
- previously tasks had 3 different statuses: `in-progress`, `done`, `cancelled`. Cancelled changed into `archived`. Now possible statuses are `in-progress`, `done`, `archived`.
- updating a task was overwriting the relations always, now it's only overwriting the relations if only they're sent
- Now we're creating suggested tasks (which are retrieved from `suggested-tasks.json`) upon tenant creation. These tasks will have no assignee, and type=`suggested`
-  `UserRepository.autocomplete` endpoint was not mapping the `fullName` field correctly, resulting in only returning the `email`. Now it's returning both `email` and `fullName` as a result of autocomplete request.
        
## Checklist ✅
- [x] Tests are passing.  


gaspergrom (approved):
Based on what frontend needs it looks good to me

PR #222, opened 2022-11-16
Reset dashboard after tenant selection

mariobalca:
# Changes proposed ✍️
- Reset dashboard after tenant selection

## Checklist ✅
- [x] Tests are passing.  
- [ ] ~Environment variables have been updated~
  - [ ] ~Front-end: `frontend/.env.dist`~
  - [ ] ~Backend: `backend/.env.dist`, `backend/.env.dist.staging`, `backend/.env.dist.staging`.~
  - [ ] ~[Configuration docs](https://docs.crowd.dev/docs/configuration) have been updated.~
  - [ ] ~Team members only: update environment variables in Password manager and update the team~

### DIFF
--- a/frontend/src/modules/dashboard/components/dashboard-task.vue
+++ b/frontend/src/modules/dashboard/components/dashboard-task.vue
@@ -202,6 +202,10 @@ const taskCreatePermission = computed(
 )
 
 const storeUnsubscribe = store.subscribeAction((action) => {
+  if (action.type === 'auth/doRefreshCurrentUser') {
+    fetchTasks()
+    fetchSuggestedTasks()
+  }
   if (action.type === 'task/reloadOpenTasks') {
     fetchTasks()
   }
--- a/frontend/src/modules/dashboard/store/actions.js
+++ b/frontend/src/modules/dashboard/store/actions.js
@@ -5,6 +5,13 @@ import { ConversationService } from '@/modules/conversation/conversation-service
 import moment from 'moment'
 
 export default {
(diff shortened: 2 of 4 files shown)


frontend/src/modules/dashboard/store/actions.js
```diff
@@ -5,6 +5,13 @@ import { ConversationService } from '@/modules/conversation/conversation-service
 import moment from 'moment'
 
 export default {
+  async reset({ dispatch }) {
+    dispatch('setFilter', {
```
joanagmaia, line 9: I think it should be `setFilters` instead of `setFilter` right?
At least according to the available action on the same file. I'm also getting a console error for the unknown action. Could you please confirm if it's ok? 🙏

frontend/src/modules/dashboard/components/dashboard-task.vue
```diff
@@ -202,6 +202,10 @@ const taskCreatePermission = computed(
 )
 
 const storeUnsubscribe = store.subscribeAction((action) => {
```
joanagmaia, line 204: Sorry if I'm lacking context but should the tasks store be handled the same way as integrations? What I mean is that integrations are fetched from the start of the application regardless of opening the Integrations page. Once the tenant switches, the integrations are reset. Shouldn't it be the same for tasks?

Nonetheless, I think I found an issue with this approach which is the number of uncompleted tasks next to the "Tasks" menu item. After switching tenant, that value is only shown once you access the Tasks page. Could you please confirm if it's an issue? 🙏

PR #226, opened 2022-11-16
Fix delete task color in dropdown

mariobalca:
# Changes proposed ✍️
- Fix delete task color in dropdown 
        
## Checklist ✅
- [x] Tests are passing.  
- [ ] ~Environment variables have been updated~
  - [ ] ~Front-end: `frontend/.env.dist`~
  - [ ] ~Backend: `backend/.env.dist`, `backend/.env.dist.staging`, `backend/.env.dist.staging`.~
  - [ ] ~[Configuration docs](https://docs.crowd.dev/docs/configuration) have been updated.~
  - [ ] ~Team members only: update environment variables in Password manager and update the team~


frontend/src/modules/task/components/task-dropdown.vue
```diff
@@ -45,13 +45,14 @@
...
+        class="w-55 !text-red-500"
         divided
         :command="{
           action: 'taskDelete'
         }"
-        ><i class="ri-delete-bin-line mr-1 text-red" /><span
-          class="text-red"
+        ><i
+          class="ri-delete-bin-line mr-1 !text-red-500"
+        /><span class="text-red"
```
joanagmaia, line 55: To avoid having to use `!important`, maybe it would better to just replace the `text-red` class that was being used both in the `<i>` and in the `<span>`, with `text-red-500`.
I think the other app dropdowns are using that

mariobalca: You sure? I did try to look if we were doing it differently somewhere but I only found this approach 😬

joanagmaia: You can check in [`automation-dropdown.vue`](https://github.com/CrowdDotDev/crowd.dev/blob/80b9351121f106b991162bf42bb5d3d527058455/frontend/src/modules/automation/components/automation-dropdown.vue#L41) or [`activity-dropdown.vue`](https://github.com/CrowdDotDev/crowd.dev/blob/80b9351121f106b991162bf42bb5d3d527058455/frontend/src/modules/activity/components/activity-dropdown.vue#L24) for instance

But this is just a small detail, it was just to avoid using the !important. But yeah if it's not consistent throughout the app then just feel free to merge no problem

joanagmaia (approved):

PR #236, opened 2022-11-18
Oauth2 PKCE state using redis

epipav:
# Changes proposed ✍️
- Moves all passport strategies from routes to `passportStrategyMiddleware`
- New redis middleware that makes the initial connection and binds it to `req.redis`
- Twitter oauth2 strategy now saves/retrieves PKCE state from redis. Key for each user is 
  `"cache_twitterPKCE:${userId}"`. Also can support other platforms by sending another instance of `RedisCache` into the `RedisPKCEStore` constructor. Keys are short lived. (5 minutes)
- Failures on oauth flow, now api is redirecting to frontend with `error=true` in url parameters.
        
## Checklist ✅
- [x] Tests are passing.  


backend/src/middlewares/redisMiddleware.ts
```diff
@@ -0,0 +1,18 @@
+import { getServiceLogger } from '../utils/logging'
+import { createRedisClient } from '../utils/redis'
+
+const log = getServiceLogger()
+
+export async function redisMiddleware(req, res, next) {
+  try {
+    const redis = await createRedisClient(true)
```
themarolt, line 8: Creation of redis client is an expensive operation - the way it's usually done is that only 1 or 2 instances are created at the application boot time and they are used statically. You can still inject them here in the middleware but they shouldn't be created for each request. It is also a memory leak since you have to call `quit` method once you don't need the client anymore otherwise it keeps an open connection to redis. And here we lose the reference to the client which wasn't properly disposed of.

themarolt (changes requested):

backend/src/middlewares/redisMiddleware.ts
epipav: Moved the redis connection init to the app boot with setImmediate :+1: middleware is getting this object and setting req.redis to this

PR #255, opened 2022-11-25
Add website, location, and github to organizations

mariobalca:
# Changes proposed ✍️
- Add the following columns to the `organizations` database table:
  - `website TEXT NULL`
  - `location TEXT NULL`
  - `github JSONB NULL`

**For the reviewer:** After taking a look at the `organizationService.ts` I noticed that we were using the **url** in a quite a few different places and use cases (mostly related to enrichment), so I've decided to simply add new columns to fill in the gaps we have with the mocks/frontend. Please let me know if you guys think we should not move forward like this.

## Checklist ✅
- [ ] Tests are passing.  


backend/src/serverless/integrations/webhooks/github.ts
```diff
@@ -379,6 +379,7 @@ export default class GitHubWebhook {
               location: fromAPI.location ?? null,
               logo: fromAPI.avatarUrl ?? null,
               url: fromAPI.url ?? null,
+              github: { handle: fromAPI.resourcePath },
```
epipav, line 382: shouldn't this be `fromApi.url` ? The graphql response has the gh organization link in `url`

epipav, line 382: shouldn't this be `fromApi.url` ? The graphql response has the gh organization link in `url`

epipav (changes requested):
Looking good! 
We also need to add the change in organanization parse to `githubIntegrationService`

backend/src/serverless/integrations/webhooks/github.ts
joanagmaia: At least from sample data the url is different from the handles. Handles should indicate platform profiles/handlers right? For instance, for crowd, the url could be `https://www.crowd.dev/` and the github handle should be `CrowdDotDev`

epipav (approved):

PR #318, opened 2022-12-06
Tenant plans and feature flagging

epipav:
# Changes proposed ✍️
- Updated tenant field: `plan`. It'll only accept the new plans `Growth` and `Essential`
- New tenant fields: `isTrialPlan: boolean` and `trialEndsAt: string`
- Posthog is initialized on app boot, it will be available in `req.posthog` for the requests
- On `tenant.create` and `tenant.find` API calls, we're now sending tenant information with plan and # of existing automations to Posthog
- `automation.create` endpoint now checks feature flag `automations`. This feature flag will return false if `plan == Essential` and` # of automations > 1`
- `automation.create` and `automation.destroy` endpoints now also send tenant properties to posthog.
- Community help center `settings.update` endpoint now checks feature flag `community-help-center-pro`, which is only enabled in Growth plan.
- util function `isFeatureEnabled` will return `true` for self hosted editions.
- new util npm script `script:change-tenant-plan`: can be used to switch a tenant's plan, also sends the information to posthog
- new cron job: `downgradeExpiredTrialPlans` will check tenants that have an expired trial plan. If found downgrades the plan and sends the data to posthog
- Reverse trial: on tenant creation now we set the `plan=Growth`, `isTrialPlan=true` and also trial expiry time. Trial expiry time will be `2023-01-15` until 1st of January. Then trial time will be calculated as `now  + 2 weeks`
  

## Checklist ✅
- [x] Tests are passing.
- [x] Environment variables have been updated:
  - [x] Local frontend configuration: `frontend/.env.dist.local`, `frontend/.env.dist.composed`.
  - [x] Local backend: `backend/.env.dist.local`, `backend/.env.dist.composed`.
  - [x] [Configuration docs](https://docs.crowd.dev/docs/configuration) have been updated.


backend/src/api/automation/automationCreate.ts
```diff
@@ -19,9 +24,25 @@ import track from '../../segment/track'
...
+
   const payload = await new AutomationService(req).create(req.body.data)
 
   track('Automation Created', { ...payload }, { ...req })
 
+  identifyTenant(req)
+
+  // wait a small window for posthog
+  // to process the queue message before returing back
+  await timeout(100)
```
themarolt, line 45: Is this absolutely necessary? I really don't like having random waiting because something else needs processing :/.

backend/src/api/automation/automationCreate.ts
```diff
@@ -19,9 +24,25 @@ import track from '../../segment/track'
  */
 export default async (req, res) => {
   new PermissionChecker(req).validateHas(Permissions.values.automationCreate)
+
+  if (!(await isFeatureEnabled(FeatureFlag.AUTOMATIONS, req.currentTenant.id, req.posthog))) {
```
themarolt, line 28: Could this be a middleware? Might be easier to use that way?

backend/src/feature-flags/isFeatureEnabled.ts
```diff
@@ -0,0 +1,18 @@
...
+export default async (
+  featureFlag: FeatureFlag,
+  tenantId: string,
+  posthog: PostHog,
+): Promise<boolean> => {
+  if (API_CONFIG.edition === Edition.COMMUNITY) {
+    return true
+  }
+
+  const featureFlagEnabled = await posthog.isFeatureEnabled(featureFlag, '', {
```
themarolt, line 14: Can certain feature flags be cached? Like for an hour or a day or depending on the conditions behind it? I imagine this being used in almost CRUD all API requests and could be slowing things down.

backend/src/api/automation/automationCreate.ts
epipav: Changed this into a middleware for api endpoints.
Only currently this endpoint is using it as a middleware tho - The other example is in community help center pro, when customUrl exists, I think we can keep it as inline since when customUrl is not existent, we don't really need to call the feature flag

backend/src/feature-flags/isFeatureEnabled.ts
epipav: Posthog is already caching it to memory, so consequent isFeatureEnabled calls will be retrieved from memory.
However I don't think there's an easy way to use this effectively because of multiple api pods.
Any ideas?

PR #392, opened 2023-01-04
Improved copy for widget tooltips

epipav:
# Changes proposed ✍️
- Changes the tooltip of the widgets without dimensions from format `[Cube] Count: {value}` to `{value} [Plural or Singular Cube Name]`
eg: Changes `[Activities] Count: 5` to `5 Activities`

  
- ### Screenshots (front-end changes only)
tooltip-update

## Checklist ✅
- [ ] Tests are passing.
- [x] Environment variables have been updated:

### DIFF
--- a/frontend/src/i18n/en.js
+++ b/frontend/src/i18n/en.js
@@ -510,22 +510,10 @@ const en = {
   widget: {
     cubejs: {
       tooltip: {
-        Activities: {
-          singular: 'Activity',
-          plural: 'Activities'
-        },
-        Members: {
-          singular: 'Member',
-          plural: 'Members'
-        },
-        Conversations: {
-          singular: 'Conversation',
-          plural: 'Conversations'
-        },
-        Organizations: {
-          singular: 'Organization',
-          plural: 'Organizations'
-        }
+        Activities: 'Activity',
+        Members: 'Member',
+        Conversations: 'Conversation',
+        Organizations: 'Organization'
       },
       cubes: {
         Activities: 'Activities',
--- a/frontend/src/i18n/en.js
+++ b/frontend/src/i18n/en.js
@@ -509,11 +509,29 @@ const en = {
 
   widget: {
     cubejs: {
+      tooltip: {
+        Activities: {
+          singular: 'Activity',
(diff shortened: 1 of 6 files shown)


frontend/src/modules/report/report-charts.js
```diff
@@ -13,6 +15,30 @@ const defaultChartOptions = {
     '#06B6D4',
     '#F97316'
   ],
+  library: {
```
joanagmaia, line 18: By changing this here, some chart types are overriding it later on the file. For example donut/pie graph has a specific configuration for plugins.
Maybe add this configuration to the bottom of the file where we are returning
```
return {
    ...defaultChartOptions,
    ...chartTypeOptions
  }
```

If you check the current default report in the "Activities by Type" donut chart, you can still see the old labels

frontend/src/i18n/en.js
```diff
@@ -509,11 +509,29 @@ const en = {
 
   widget: {
     cubejs: {
+      tooltip: {
+        Activities: {
+          singular: 'Activity',
```
joanagmaia, line 514: Instead of having to add the singular/plural logic you can just set the singular format, for example: `Activities: 'Activity'`, and then when getting the translation you could use `pluralize` (we are already using in some places, automation-list-page.vue):
```
callbacks: {
          label: (context) => {
            return pluralize(
              i18n(`widget.cubejs.tooltip.${context.dataset.label}`),
              context.dataset.data[context.dataIndex],
              true
            )
          }
        }
```
This will already do all the logic to render it as:
1 Member
2 Members

frontend/src/modules/report/report-charts.js
epipav: Updated this part :+1:  When there's a dimension, we'll still use old style tooltips (like dimension: value). Graphs without dimensions will use the new style tooltips

frontend/src/i18n/en.js
epipav: Also updated - we're using pluralize now and only keeping the singular form in translations

joanagmaia (approved):
LGTM! Thanks for taking care of this

PR #395, opened 2023-01-05
Fixed hashtags not being saved

joanreyero:
# Changes proposed ✍️
- Twitter hashtags are now saved. I added them to the redirect URL so they are not lost.

## Checklist ✅
- [x] Tests are passing.

### DIFF
--- a/backend/src/api/integration/helpers/twitterAuthenticate.ts
+++ b/backend/src/api/integration/helpers/twitterAuthenticate.ts
@@ -9,8 +9,7 @@ export default async (req, res, next) => {
 
   const state = {
     tenantId: req.params.tenantId,
-    redirectUrl: req.query.redirectUrl,
-    hashtags: req.query.hashtags,
+    redirectUrl: `${req.query.redirectUrl}&hashtags=${req.query.hashtags}`,
     crowdToken: req.query.crowdToken,
     platform: PlatformType.TWITTER,
     userId: req.currentUser.id,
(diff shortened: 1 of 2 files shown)


backend/src/api/integration/helpers/twitterAuthenticate.ts
```diff
@@ -9,8 +9,7 @@ export default async (req, res, next) => {
 
   const state = {
     tenantId: req.params.tenantId,
-    redirectUrl: req.query.redirectUrl,
-    hashtags: req.query.hashtags,
```
epipav, line 13: I'd say let's keep the hashtags in the state instead of redirect url - so the fields required are separated 
The reason hashtags were not being forwarded back to authenticate callback is it's missing in `redisPKCEStore` line 49. After adding the hashtags there it should forward successfully

backend/src/api/integration/helpers/twitterAuthenticateCallback.ts
```diff
@@ -5,10 +5,10 @@ import IntegrationService from '../../../services/integrationService'
 export default async (req, res) => {
   // Checking we have permision to edit the integration
   new PermissionChecker(req).validateHas(Permissions.values.integrationEdit)
-
   const urlSearchParams = new URLSearchParams(req.query.state)
   const redirectUrl = urlSearchParams.get('redirectUrl')
-  const hashtags = urlSearchParams.get('hashtags')
```
epipav, line 11: Also, we can keep these because it'll be forwarded back in the state parameter, but since it's serialized and deserialized, we need to update the below part as 
```
  const integrationData = {
    profileId: req.user.twitter.profile.id,
    token: req.user.twitter.accessToken,
    refreshToken: req.user.twitter.refreshToken,
    hashtags: hashtags.split(","),
  }
```

epipav (changes requested):
I think we can still use the state to forward the variables to the callback (like we're doing for userId, token or profileId)

PR #463, opened 2023-01-27
Fix some unhandled report issues

joanagmaia:
# Changes proposed ✍️
- Fix the following errors thrown in log rocket:
   - [TypeError: Cannot set properties of undefined (setting 'loading')](https://app.logrocket.com/nm6fil/crowddev/issue/290067027?issueType=EXCEPTION&referrer=slack)
   - [TypeError: e.byId[t] is undefined](https://app.logrocket.com/nm6fil/crowddev/issue/287773229?issueType=EXCEPTION&referrer=slack)
   - [TypeError: Cannot read properties of undefined (reading 'name')](https://app.logrocket.com/nm6fil/crowddev/issue/287635572?issueType=EXCEPTION&referrer=slack)
   - [TypeError Cannot read properties of undefined (reading 'some')](https://app.logrocket.com/nm6fil/crowddev/issue/i?from=issues%2Fneeds-triage&issueFilters=%255B%257B%2522exceptionGroupID%2522%253A%257B%2522operator%2522%253A%2522EQ%2522%252C%2522strings%2522%253A%255B%2522248884297%2522%255D%257D%257D%255D&issueType=EXCEPTION)

## Checklist ✅
- [ ] Tests are passing.

### DIFF
--- a/frontend/src/modules/layout/components/workspace-dropdown.vue
+++ b/frontend/src/modules/layout/components/workspace-dropdown.vue
@@ -188,7 +188,7 @@ const clickOutsideListener = (event) => {
       // we need the following condition to validate clicks
       // on popovers that are not DOM children of this component,
       // since popper is adding fixed components to the body directly
-      event.path.some(
+      event.path?.some(
         (o) => o.className?.includes('el-popper') || false
       )
     )
--- a/frontend/src/modules/report/pages/report-list-page.vue
+++ b/frontend/src/modules/report/pages/report-list-page.vue
@@ -17,42 +17,49 @@
       </div>
     </div>
 
-    <!-- Template reports -->
-    <div v-if="computedTemplates.length">
-      <div
-        class="text-gray-900 font-semibold text-base mb-6"
-      >
-        Default reports
-      </div>
+    <div
+      v-if="loading"
+      v-loading="loading"
+      class="app-page-spinner h-16 !relative !min-h-5"
+    ></div>
+    <div v-else>
+      <!-- Template reports -->
+      <div v-if="computedTemplates.length">
+        <div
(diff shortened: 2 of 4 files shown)


mariobalca (commented):
Changes look good to me, but I'm wondering why didn't/doesn't the app redirect the user to the `/404` page after validating that the report does not exist (so after getting the 404 from the API) 🤔 
Did you understand what was happening?

joanagmaia:
> Changes look good to me, but I'm wondering why didn't/doesn't the app redirect the user to the `/404` page after validating that the report does not exist (so after getting the 404 from the API) 🤔 Did you understand what was happening?

Thanks for raising that point!
I've actually update the PR overall description because some the errors there weren't fixed by this PR (sorry about that)

Nonetheless, from what I've checked and saw the current behaviour is if you try to access a report that doesn't exist you get redirected to the reports list page (I think this is also ok instead of redirecting to a 404, what do you say?). However, the problem is that in the reports view page components we were only protecting for the loading and not error state. So what I think was happening was that after the loading state and before the error state, a lot of properties were being used on a report object that did not exist. By preventing the page content from being rendered and accessed in an error scenario, we are preventing these undefined errors

mariobalca:
I see, thanks for clarifying!

>the current behaviour is if you try to access a report that doesn't exist you get redirected to the reports list page (I think this is also ok instead of redirecting to a 404, what do you say?)

You're right, I was not aware that this was how we're handling it at the moment. Not a big problem for me — ultimately I guess this might add less frustration to the user (since he won't be seeing the 404 page), but I'm not sure if this is "standard behaviour", maybe we should check it later on with Nuno.

mariobalca (approved):

PR #572, opened 2023-02-24
Allow to hide/show attributes

joanagmaia:
# Changes proposed ✍️
- Add show/hide button next to attribute's field -> Only render the button if attribute cannot be deleted
- Once the button is clicked, the user is prompted with a message to confirm their action. Once the action is confirmed, the attribute will either be shown/hidden from member's profiles
- Fix Attributes section in New members -> The filtering for custom attributes was throwing an error for new member form, because there was no `record`. For new members form it will only be shown attributes that can be deleted. For edit form, all member's attributes will be displayed
- Refactor `get-custom-attribuetes` so that `show` property can be considered or not when filtering depending on the use case. For members filters (considered), for forms (not considered)
  
### Screenshots (front-end changes only)
Screenshot 2023-02-24 at 11 57 17

## Checklist ✅
- [ ] Tests are passing.

### DIFF
--- a/frontend/src/modules/member/components/form/member-form-attributes.vue
+++ b/frontend/src/modules/member/components/form/member-form-attributes.vue
@@ -12,15 +12,13 @@
       >
     </div>
     <div :class="showHeader ? 'col-span-2' : 'col-span-3'">
-      <div
-        class="grid grid-cols-12 gap-3 border-b h-8 items-center"
-      >
+      <div class="flex gap-3 border-b h-8 items-center">
         <span
-          class="uppercase text-gray-400 text-2xs font-semibold tracking-wide col-span-4"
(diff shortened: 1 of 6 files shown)


joanreyero (commented):
The code and functionality look great! 

A couple of suggested changes:

- I think the eye icon is mixed up. Notion has the "on" eye when the attribute is visible and the off eye when it is not. I think that makes more sense. When I visited a member for the first time, it looked like all attributes were hidden.

Screenshot 2023-02-24 at 14 17 56

- I would grey out the icon when it is hidden, to give more visibility (see the notion screenshot).

- I would make the icons a bit smaller, and the default a bit lighter, but this is more personal preference 😀

gaspergrom (approved):

PR #587, opened 2023-02-28
Add limit to autocomplete member's dropdown

joanagmaia:
# Changes proposed ✍️
- Remove deprecated components: `platform-autocomplete-input.vue`
- Remove `member-autocomplete-input` component to simplify it's purpose. Directly use `autocomplete-one-input` component. Create/add `totalFn` method to `autocomplete-one-input` to fetch the total number of resources within the dropdown (autocomplete endpoint does not gives us this information)
- Simplify `tag-autocomplete-input` as it was not using some of its components/data
- Fix: Clear member to merge when merge member dialog opens
- Refactor `autocomplete-one-input` to load results with lazy loading

## Checklist ✅
- [ ] Tests are passing.


frontend/src/modules/member/components/member-selection-dropdown.vue
```diff
@@ -60,6 +62,16 @@ const fetchFn = async (query, limit) => {
...
+
+const totalFn = async () => {
+  try {
+    const response = await MemberService.list({}, '', 1, 0)
+
+    return response.count
+  } catch (e) {
+    return null
+  }
+}
```
joanreyero, line 74: This is only called once, no?

frontend/src/shared/form/autocomplete-one-input.vue
```diff
@@ -183,9 +227,23 @@ export default {
...
     },
 
+    async onScroll() {
+      // If limit is higher or equal to total number of members, do not fetch for more options
+      if (this.limit >= this.total) {
+        return
+      }
+
+      this.limit =
+        this.limit + AUTOCOMPLETE_SERVER_FETCH_SIZE
```
joanreyero, line 237: One question, it looks to me like we are increasing the limit instead of keeping it constant, but using pagination. Is that correct? 

If so, why? 😀

joanreyero (commented):
Thanks for handling this so quickly! Added a couple of questions

frontend/src/modules/member/components/member-selection-dropdown.vue
joanagmaia: Yes this is only called in the `created` lifecycle of the `autocomplete-one-input` component, so it will only be called the first time the component is rendered

frontend/src/shared/form/autocomplete-one-input.vue
joanagmaia: Yes that's what I'm doing. Unfortunately, from my understanding,  the `member/autocomplete` endpoint that we are using for this component only handles the `limit` parameter, not the `offset`. It also only retrieves the results and not the pagination. So I had to handle it this way
Does it not work like this?

joanagmaia: So right now we have lazy loading in the dropdown. We start with a limit of 20, and as the user scrolls down we just increase the limit

joanagmaia:
@joanreyero just updated the PR so that we have a limit of 20 results in the dropdown

PR #603, opened 2023-03-03
Search filter initial value setting

gaspergrom:
# Changes proposed ✍️
- search initial value setting when returning to activities, members or organizations view

## Checklist ✅
- [ ] Tests are passing.

### DIFF
--- a/frontend/src/shared/filter/components/type/filter-type-search.vue
+++ b/frontend/src/shared/filter/components/type/filter-type-search.vue
@@ -15,12 +15,13 @@
 </template>
 <script setup>
 import {
-  h,
+  computed,
   defineEmits,
   defineProps,
-  watch,
+  h,
   ref,
-  computed
+  watch,
+  onMounted
 } from 'vue'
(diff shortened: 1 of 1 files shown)


frontend/src/shared/filter/components/type/filter-type-search.vue
```diff
@@ -76,4 +77,10 @@ const debouncedChange = debounce((value) => {
     value
   })
 }, 300)
+
+onMounted(() => {
```
joanagmaia, line 81: We also have the edge case between active views, so whenever an active view changes, the search bar is also being cleaned but the results are not
https://www.loom.com/share/0416534c05fc410d8ea4da0bb5321ca2

I'm actually not sure what would be the best UX here, currently the search query is being stored per view in the store. If we want to keep it that way, than I think that's the only thing missing to fix regarding the search

gaspergrom: fixed

joanagmaia (approved):

PR #607, opened 2023-03-03
New onboarding

gaspergrom:
# Changes proposed ✍️
- new workspace setup
- removal of integration onboarding step
- fix copy of banner
- onboarding guides
- eagle eye modal
- delete sample data after integration is connected
  
- ### Screenshots (front-end changes only)
image
image

## Checklist ✅
- [ ] Tests are passing.


epipav:
Backend changes:
- new member field: `enrichedBy: array`
- new report field: `viewedBy: array`
- new tenantUser field `invitedBy: uuid`
- moved eagle eye settings to tenantUser level instead of user
- new quickstart endpoints for getting tenantUser specific quickstart guides
- quickstart settings (also stored in tenantUser level) 
- we delete sample data on onboarding integration processing now
- since we now store eagle eye related settings in tenantUser level, we can successfully forward tenant id for eagle eye email digest, that'll fire correct events for tenant-user pair via sendgrid webhooks

frontend/src/modules/dashboard/components/dashboard-guides.vue
```diff
@@ -0,0 +1,169 @@
...
+}
+
+const dismissGuides = () => {
+  ConfirmDialog({
+    type: 'info',
+    title:
+      'Do you really want to dismiss our Quickstart Guide?',
+    message:
+      'Users that follow our Quickstart Guide are 73% more likely to successfully set-up crowd.dev.',
+    icon: 'ri-information-line',
```
joanagmaia, line 141: Should be blue and not grey, could you check if we have that configuration

joanagmaia (commented):
Just a couple of things I noticed:
- I couldn't comment because it was an image file, but we already had the same banner image for eagle-eye. It's named `banner` and is in the same directory. Could you use that one?
- The copy of the quick start guides don't seem to match the design (don't know if something changed). E.g. Set Eagle eye -> Discover content in your niche. Also the video links are all redirecting to the eagle eye video. But I think both of these issues are related to backend right?
- The main issue I found was the way that quick start guides are being rendered. They take a lot of time to appear on the screen and since there is no loading it just pops up after some time. I think this is because you are also waiting for the videos to load, maybe they could only be loaded when the step is opened. Also, since we are always calling the endpoint on dashboard mount, even if you've visited the dashboard before and if you go to one page and then get back to dashboard you need to wait again for them to be loaded, and I think it would be expected for the guide to already be there. Should we make the first call right from the app start, and then just make another call on dashboard mount to see if any of the steps were updated in the meantime? Here's a quick video of an example of what I'm traying to say https://www.loom.com/share/47d55f57fd6f420ba82a2792eedacbf6

gaspergrom:
> 

1. I checked before and its different image, although cards are the same, there is cropped on left side and more cards visible on right with the background
2. Already fixed texts, we probably had wrong text at the begining
3. Will see optimal way to handle this

joanagmaia (approved):

themarolt (approved):

PR #616, opened 2023-03-06
Alerting for faulty integrations

joanreyero:
# Changes proposed ✍️
- Added an alerting system when integrations are not getting data. We currently have 3 alerts:
  - one for general integrations, where if there is no data after 2 hours, the integration goes into no data status (also displayed in the frontend) and we get a Slack alert
  - one for Slack and Discord, where if we don't get any messages activity after 2 hours does the same as above
  - one where if an integration does not have new data after 3 days, we get an alert (but not the user).

If the integration goes into error state when onboarding, we also get an alert.

## Checklist ✅
- [ ] Tests are passing.


backend/src/serverless/microservices/nodejs/integration-data-checker/integrationDataCheckerWorker.ts
```diff
@@ -0,0 +1,122 @@
...
+  userContext: IRepositoryOptions,
+) {
+  return sendSlackAlert(SlackAlertTypes.DATA_CHECKER, integration, userContext, log, settings)
+}
+
+function generateDate(timeframe) {
+  const now = moment()
+  if (timeframe.includes('hour')) {
+    const hours = parseInt(timeframe, 10)
+    return now.subtract(hours, 'hours')
```
epipav, line 112: NIT:
I was confused here because we're sending an example string '2 hours' into `parseInt`. 
I didn't know parseInt was ignoring all the characters after whitespace. Can we split this string with whitespace and send/check split values for clarity?

backend/src/utils/slackAlerts.ts
```diff
@@ -0,0 +1,131 @@
...
+) {
+  const blocks = getBlocks(alertType, integration, userContext, log, settings)
+  const url = SLACK_ALERTING_CONFIG.url
+  await axios.post(url, blocks)
+}
+
+function getBlocks(alertType, integration, userContext, log, settings) {
+  const tenantName = userContext.currentTenant.name
+  const isPayingCustomer = userContext.currentTenant.plan
+  const isTrial = userContext.currentTenant.trialEndsAt > new Date()
```
epipav, line 27: NIT:
I think we have `currentTenant.isTrialPlan` field available

epipav (commented):
Looking good :fire:  I included few comments

PR #793, opened 2023-04-21
Bugfix/member upsert doesn't always work when called from integrations

themarolt:
# Changes proposed ✍️

### What
- sometimes ActivityService.createWithMember was called with the incorrectly formatted activity or member or both
- MemberService.update and MemberService.upsert when called with an existing identity in some cases didn't correctly propagate the changes to the database - it tried to insert the same identity for a member that already had it

## Checklist ✅
- [ ] ~Add screehshots to the PR description for relevant FE changes~

### DIFF
--- a/backend/src/database/repositories/memberRepository.ts
+++ b/backend/src/database/repositories/memberRepository.ts
@@ -80,19 +80,21 @@ class MemberRepository {
     `
 
     for (const platform of Object.keys(username) as PlatformType[]) {
-      const identity = username[platform]
-      await seq.query(query, {
-        replacements: {
-          memberId: record.id,
-          platform,
-          username: identity.username,
-          sourceId: identity.sourceId || null,
-          integrationId: identity.integrationId || null,
-          tenantId: tenant.id,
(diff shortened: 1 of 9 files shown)


backend/src/database/repositories/memberRepository.ts
```diff
@@ -374,15 +372,24 @@ class MemberRepository {
...
+        and mi.username in (:usernames)
     `
 
+    const usernames: string[] = []
+    if (typeof username === 'string') {
+      usernames.push(username)
+    } else if (Array.isArray(username)) {
+      usernames.push(...username)
+    } else {
+      throw new Error('Unknown username format!')
```
joanreyero, line 384: Can we return a list of the formats we allow?

joanreyero (approved):
Just a little nit

backend/src/database/repositories/memberRepository.ts
themarolt: Yes - will add in the error message.

PR #855, opened 2023-05-12
Soft-delete integrations when removing them

sausage-todd:
# Changes proposed ✍️

### <samp>🤖 Generated by Copilot at 2d0fa4d</samp>

Removed `force` option from `sync` method in `integrationRepository.ts` to prevent data loss. This is part of a pull request to improve the database configuration and migrations for the backend.
​

### <samp>🤖 Generated by Copilot at 2d0fa4d</samp>

> _`force` option gone_
> _`sync` method updates table_
> _data stays in fall_

### Why
To make troubleshooting/debugging of integrations easier

### <samp>🤖 Generated by Copilot at 2d0fa4d</samp>

* Remove `force` option from `sync` method of `IntegrationRepository` class to prevent table dropping and data loss ([link](https://github.com/CrowdDotDev/crowd.dev/pull/855/files?diff=unified&w=0#diff-5b1c50661ca3f395a4ec888993b46eadd35b277a777e88e0e81e1f88461124a3L116))

## Checklist ✅

### DIFF
--- a/backend/src/database/repositories/integrationRepository.ts
+++ b/backend/src/database/repositories/integrationRepository.ts
@@ -113,7 +113,6 @@ class IntegrationRepository {
 
     await record.destroy({
       transaction,
-      force: true,
     })
 
     await this._createAuditLog(AuditLogRepository.DELETE, record, record, options)


joanreyero (commented):
Congrats on your first PR! 🚀

Curious about what will happen if we: 
- connect an integration
- disconnect
- connect again

Could you send this to staging and try it out?

sausage-todd:
These are very good questions, let me see

sausage-todd:
Ok, so here's how it works:
1. When connecting again an integration that was previously disconnected, we create a new entry in the database, while keeping the soft-deleted on untouched.
    - I believe this is a good approach, because it should make troubleshooting easier for us by keeping potentially broken data intact.
2. `nodejs-worker` doesn't pick up soft-deleted integrations
    - I tried it both on staging and locally. According to logs, it doesn't start processing soft-deleted integrations
    - It makes sense, because `nodejs-worker` uses [`IntegrationRepository.findAllActive()`](https://github.com/CrowdDotDev/crowd.dev/blob/6f747d8b0b2f3810b4c44074c03feec06854ae1d/backend/src/database/repositories/integrationRepository.ts#L150C2-L167) for retrieving integrations to be processed, and it still uses `Sequelize`.
3. Soft-deleted integrations are ignored everywhere else as well
    - As long as we don't explicitly fetch soft-deleted rows with Sequelize
    - There is [only one place that uses raw sql to access integrations](https://github.com/CrowdDotDev/crowd.dev/blob/6f747d8b0b2f3810b4c44074c03feec06854ae1d/backend/src/database/repositories/integrationRepository.ts#L169C5-L193), and it shouldn't be a problem, because it's used to fix integrations stuck in-progress

cc @joanreyero @themarolt

themarolt (approved):

PR #967, opened 2023-06-08
"+ Add Activity Type should be available even if there are no custom activities"

### LINKED ISSUE #964: [C-1529] "+ Add Activity Type" button should show even if there are no custom activities
### crowd.dev edition

Cloud (hosted)

### Version

_No response_

### Link

https://app.crowd.dev/activities

### Describe the problem

# Steps to reproduce

1. Go to Activities page.
2. Click on Add Activity button on top right
3. Now if you don't have any custom activities already created than "Activity Type" dropdown shows no data and user cannot create a new custom activity from there

## Actual
Screenshot from 2023-06-08 12-12-09

## Expected
Screenshot from 2023-06-08 12-19-11

 

### Describe the improvement

Instead it should show "+ Add Activity" button even if there are no custom activities

### Additional context

_No response_

<sub>[C-1529](https://linear.app/crowddotdev/issue/C-1529/add-activity-type-button-should-show-even-if-there-are-no-custom)</sub>

Dhruwang:
# Changes proposed ✍️
Closes #964 
### What

On Activities -> Add activity
Activity type dropdown should allow user to add new activity type even if there are not custom activity presnt
Made "+ Add Activity Type" button available
​

https://github.com/CrowdDotDev/crowd.dev/assets/67850763/9783ca6a-5e35-47ed-a17f-9d88bb058f96

### <samp>🤖 Generated by Copilot at d672506</samp>

*  Changed `+ Add activity type` option to an el-option element and renamed types.custom to types.custom.other in `activity-form-drawer.vue` ([link](https://github.com/CrowdDotDev/crowd.dev/pull/967/files?diff=unified&w=0#diff-277ffbe89d66e672128e3830c545d954d516d187f7e15e51374f63a6892806b0L89-R97))


gaspergrom (approved):

frontend/src/modules/activity/components/activity-form-drawer.vue
```diff
@@ -86,14 +86,15 @@
...
+              :value="null"
               class="px-3 py-2.5 text-brand-500 text-xs leading-5 transition hover:bg-gray-50 cursor-pointer"
               @click="emit('add-activity-type')"
             >
               + Add activity type
-            </div>
+            </el-option>
             <div
-              v-if="Object.keys(types.custom).length > 0"
+              v-if="Object.keys(types.custom.other).length > 0"
```
joanagmaia, line 97: Hey! Could you please remove `other`? We are listing all `custom` types
Also this breaks the dropdown if the user has never created any custom activity type, because `other` won't exist

joanagmaia (changes requested):

frontend/src/modules/activity/components/activity-form-drawer.vue
Dhruwang: Actually when a user has never created a custom activity then its `types.custom` object would look something like `{}`
So calling `types.custom.other` breaks the dropdown
But when a user has created some custom activity then deleted it then its types.custom will look like this { "other": {} } 
So if we calculate length of `types.custom` then it will 1 because `types.custom = { "other": {} } `
and hence `Object.keys(types.custom).length > 0` will become true and hence Custom will be shown even if there are no custom events 
Screenshot from 2023-06-10 01-26-45
So adding a ternary opeator : `types.custom.other? Object.keys(types.custom.other).length > 0 : Object.keys(types.custom).length > 0`

joanagmaia (approved):
LGTM 🚀

PR #1001, opened 2023-06-19
Querying members with opensearch

epipav:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


services/libs/opensearch/src/client.ts
```diff
@@ -0,0 +1,26 @@
...
+          }),
+        }),
+      })
+    }
+    return new Client({
+      node: config.node,
+    })
+  }
+
+  return null
```
themarolt, line 25: might be better to throw an error here

services/scripts/build_libs.sh
```diff
@@ -16,6 +16,7 @@ LIBRARIES=(
   "conversations"
   "integrations"
   "alerting"
+  "opensearch"
```
themarolt, line 19: Also add it to watch_libs.sh script. It's a map with <lib-name>: @crowd dependencies list of that lib so that the script knows what to rebuild. In you case it should be just: opensearch: types because you only depend on @crowd/types library.

PR #1027, opened 2023-06-26
Added necessary environment variables

Bharaththop:
# Resolved compilation error

### What
server port listening error   ,fixed  through simple change

## Checklist ✅
-  New backend functionality has been unit-tested.


joanreyero:
Hey @Bharaththop! Thanks for the pull request. What was the error that caused you to open this?

Bharaththop:
> Hey @Bharaththop! Thanks for the pull request. What was the error that caused you to open this?

Yeah, If we go through the flow of commits which I raised for the pull request :
1.  Added variables to Docker files which are necessary for the image to run.
2. And in this file  [(scripts/scaffold/nginx/templates/default.conf.template)](url) changed  script as compiler not recognising ssl certificate .

backend/src/database/Dockerfile.flyway.repair
```diff
@@ -8,7 +8,7 @@ RUN apk update \
 
 USER flyway
 
-COPY ./flyway_repair.sh /repair.sh
+COPY .flyway_repair.sh /repair.sh
```
themarolt, line 11: this is not gonna work - there is no `.flyway_repair.sh` file

backend/tsconfi.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
+    "resolveJsonModule": true,
```
themarolt, line 4: We already have a `tsconfig.json` in our backend folder - and this is also a `tsconfi.json` without a `g` so it won't be used anyway.

scripts/scaffold/db/Dockerfile
```diff
@@ -1,3 +1,4 @@
 FROM postgres:13.6-alpine
+ENV POSTGRES_PASSWORD = 'example'
```
themarolt, line 2: We are starting things without `scripts/cli` bash script which actually starts this database using docker compose file `scripts/scaffold.yaml` which has all the necessary env. variables set.

themarolt (changes requested):

PR #1033, opened 2023-06-26
Move Filters exclude to operators

gaspergrom:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/frontend/src/shared/modules/filters/components/filterTypes/BooleanFilter.vue
+++ b/frontend/src/shared/modules/filters/components/filterTypes/BooleanFilter.vue
@@ -1,6 +1,5 @@
 <template>
   <div v-if="form">
-    <cr-filter-include-switch v-if="!props.hideIncludeSwitch" v-model="form.include" />
     <div class="p-4">
       <p class="text-xs text-gray-500 pb-3">
         {{ props.config.label }}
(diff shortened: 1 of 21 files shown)


frontend/src/shared/modules/filters/components/filterTypes/BooleanFilter.vue
```diff
@@ -1,6 +1,5 @@
 <template>
   <div v-if="form">
-    <cr-filter-include-switch v-if="!props.hideIncludeSwitch" v-model="form.include" />
```
joanagmaia, line 3: I think here you missed to delete the component import?
Also to remove `include` from `defaultForm`

frontend/src/shared/modules/filters/components/filterTypes/DateFilter.vue
```diff
@@ -65,16 +62,18 @@ const rules: any = {
   },
 };
 
+const checkIfBetween = (operator) => [FilterDateOperator.BETWEEN, FilterDateOperator.NOT_BETWEEN].includes(operator);
```
joanagmaia, line 65: Should this be a computed property?

joanagmaia (commented):
Some issues I found:
1. Filters are now being loaded initially but the tab does not save. So if you refresh page in "New and Active" tab what happens is that filters are loaded correctly but the selected tab is not the correct one, it goes back to the first one
2. Didn't we use to have a "is Team Member" filter in Members filters? I think so, we don't have one now so when you go to Team members tab, you don't see the filter being applied
3. Loading in members page is still not completely functional, sometimes you see the loader but then the empty state for a few seconds before listing the table. I think you can see this if you refresh on the All members tab

frontend/src/shared/modules/filters/components/filterTypes/DateFilter.vue
gaspergrom: I compare previous value with current so i added now also computed property for faster rendering

joanagmaia (approved):

PR #1408, opened 2023-09-01
C-2200 - Init tracing with OpenTelemetry

loicsaintroch:
# Changes proposed ✍️

### <samp>🤖 Generated by Copilot at e19e7ca</samp>

This pull request adds a new `tracing` library to the `services/libs` folder. The library provides a function to create and manage service tracers using OpenTelemetry. The library also includes configuration files for TypeScript, ESLint, and Prettier.
​

### <samp>🤖 Generated by Copilot at e19e7ca</samp>

> _`tracing` library_
> _adds TypeScript linting, formatting_
> _autumn code cleanup_

### <samp>🤖 Generated by Copilot at e19e7ca</samp>

*  Add `tracing` library for creating and managing service tracers (F1-F7)

## Checklist ✅


sausage-todd:
Looks good, but shall we maybe add some usage example somewhere? To kinda start with at least one trace already

services/libs/tracing/src/tracer.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const attrs: ResourceAttributes = {
+    [SemanticResourceAttributes.SERVICE_NAME]: service,
+  }
+
+  // Try to retrieve the Kubernetes namespace from file. If found, we can assume
+  // service is running in Kubernetes so we can set other Kubernetes-related
+  // resources such as pod and deployment names.
+  try {
+    const data = fs.readFileSync('/var/run/secrets/kubernetes.io/serviceaccount/namespace')
```
sausage-todd, line 35: Good idea, but we have an env variable `KUBE_MODE` specifically for this, take a look

services/libs/logging/src/logger.ts
```diff
@@ -19,7 +15,7 @@ export const getServiceLogger = (): Logger => {
   const options = {
     name: SERVICE,
     level: LOG_LEVEL as Bunyan.LogLevel,
-    stream: IS_DEV_ENV || IS_TEST_ENV ? PRETTY_FORMAT : JSON_FORMAT,
+    stream: JSON_FORMAT,
```
sausage-todd, line 18: gif

services/apps/integration_run_worker/src/queue/index.ts
```diff
@@ -76,8 +87,14 @@ export class WorkerQueueReceiver extends SqsQueueReceiver {
           throw new Error(`Unknown message type: ${message.type}`)
       }
     } catch (err) {
-      this.log.error(err, 'Error while processing message!')
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+      })
+
+      this.log.error(addTraceToLogFields(span, { err }), 'Error while processing message!')
```
sausage-todd, line 94: Oh, we'll be forgetting about doing this a lot

services/apps/integration_run_worker/package.json
```diff
@@ -20,14 +20,16 @@
...
     "@crowd/common": "file:../../libs/common",
     "@crowd/database": "file:../../libs/database",
     "@crowd/integrations": "file:../../libs/integrations",
     "@crowd/logging": "file:../../libs/logging",
     "@crowd/redis": "file:../../libs/redis",
     "@crowd/sqs": "file:../../libs/sqs",
+    "@crowd/tracing": "file:../../libs/tracing",
     "@crowd/types": "file:../../libs/types",
-    "@crowd/alerting": "file:../../libs/alerting",
+    "@opentelemetry/api": "~1.6.0",
```
sausage-todd, line 32: You sure you need this here?

services/libs/tracing/src/tracer.ts
loicsaintroch: Still, we need access to the namespace's name.

services/apps/integration_run_worker/package.json
loicsaintroch: Yes, to have access to the `Tracer` interface and `SpanStatusCode` enum.

services/libs/tracing/src/tracer.ts
sausage-todd: Then how about we get this information in a proper way like this:
https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

Because the pod name you're getting is not really a pod name. And the deployment name is based on a convention, but can get a real one from kubernetes.

services/apps/integration_run_worker/package.json
sausage-todd: Then let's do it like here:
https://github.com/CrowdDotDev/crowd.dev/blob/ae65bf25f23e61ca1e6b3f44872080da67f7be30/services/libs/logging/src/types.ts#L1-L3

Otherwise we're just going to add this dependency everywhere

sausage-todd:
Looks good apart from a couple comments. Congrats on finishing it!

loicsaintroch:
Latest changes allow automatic instrumentation and correlation in workers, with a parent span `ProcessMessage` on each message received to link each span to a common parent trace. Example:

Screenshot 2023-09-26 at 09 58 20 (2)

services/libs/tracing/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@crowd/tracing",
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "build": "rm -rf dist && ./node_modules/.bin/tsc",
```
themarolt, line 6: Please merge `main` into this branch - we stopped with the building of libraries - also check other libraries tsconfig.json after merging - there is a difference now - two properties were removed and `@/` path mapping has also been removed as it doesn't really work without building the libraries when referencing a library that has such path mapping.

themarolt (changes requested):

services/libs/tracing/package.json
themarolt: Also build_libs.sh and watch_libs.sh scripts were removed on main.

services/libs/tracing/src/tracer.ts
```diff
@@ -0,0 +1,80 @@
...
+})
+
+async function gracefulShutdown() {
+  try {
+    await sdk.shutdown()
+    console.log('Tracing successfully finished')
+  } catch (err) {
+    console.log('Error terminating tracing', err)
+  } finally {
+    process.exit(0)
```
themarolt, line 78: Not sure if we should do `process.exit(0)` here - shouldn't we let the service handle that?

Also console.logs are not gonna be properly processed by SIGNOZ AFAIK because they don't use structured format like bunyan does (JSON). 

I also see that logging library has tracing as dependency so you can't really use logging lib here because you gonna have a cyclic dependency. Not really sure what to do here - perhaps just remove the logs?

services/libs/tracing/src/tracer.ts
```diff
@@ -0,0 +1,80 @@
...
+    [SemanticResourceAttributes.SERVICE_NAME]: service,
+  }
+
+  // Apply specific resource attributes if process is running in Kubernetes.
+  // This is not handled by AWS resource detectors.
+  if (process.env['KUBE_MODE'] == '1') {
+    attrs[SemanticResourceAttributes.K8S_POD_NAME] = process.env['HOSTNAME']
+  }
+
+  sdk = new NodeSDK({
```
themarolt, line 43: How do we pass OTEL URL and other configuration options here? I would prefer to have it explicit vs. some AWS autoconfiguration happening in the SDK.

themarolt (approved):

PR #1420, opened 2023-09-05
Fixed fetching organization data aggregates when segments are enabled

themarolt:
# Changes proposed ✍️

### <samp>🤖 Generated by Copilot at e65d8ad</samp>

This pull request adds support for segment-based organization data in the backend. It enables the `organizationFind` endpoint to filter organizations by `segmentId` if the feature flag is on. It also modifies the `OrganizationRepository` and `OrganizationService` classes to use the `segmentId` parameter to calculate segment-specific metrics.
​

### <samp>🤖 Generated by Copilot at e65d8ad</samp>

> _To find an organization by id_
> _We added a parameter to `findById`_
> _It takes a `segmentId` to filter_
> _The metrics and data we deliver_
> _If the segments feature is applied_

### <samp>🤖 Generated by Copilot at e65d8ad</samp>

*  Add logic to handle segment id query parameter in `organizationFind` endpoint ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-3cd5dbd113160bb98f2226646e488303854eca68744dc1450f78332f99908d50L1-R5), [link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-3cd5dbd113160bb98f2226646e488303854eca68744dc1450f78332f99908d50L23-R38))
  * Import `isFeatureEnabled` and `FeatureFlag` from `@/feature-flags` module ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-3cd5dbd113160bb98f2226646e488303854eca68744dc1450f78332f99908d50L1-R5))
  * Check if segments feature is enabled and return error if not ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-3cd5dbd113160bb98f2226646e488303854eca68744dc1450f78332f99908d50L23-R38))
  * Pass segment id to `OrganizationService.findById` method ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-3cd5dbd113160bb98f2226646e488303854eca68744dc1450f78332f99908d50L23-R38))
* Add logic to filter and aggregate organization data by segment id in `OrganizationRepository.findById` method ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-a52bcdecd97974bf17c43be28a61e4aeb39bfc9b38ff85ef4f9769279499b7ccL449-R449), [link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-a52bcdecd97974bf17c43be28a61e4aeb39bfc9b38ff85ef4f9769279499b7ccL455-R587))
  * Add segment id parameter to method signature ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-a52bcdecd97974bf17c43be28a61e4aeb39bfc9b38ff85ef4f9769279499b7ccL449-R449))
  * Modify SQL query to use different logic depending on segment id ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-a52bcdecd97974bf17c43be28a61e4aeb39bfc9b38ff85ef4f9769279499b7ccL455-R587))
    * Use CTEs to join organization data with segment data and aggregate metrics for specific segment ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-a52bcdecd97974bf17c43be28a61e4aeb39bfc9b38ff85ef4f9769279499b7ccL455-R587))
    * Use original logic to join organization data with segment data and aggregate metrics for all segments ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-a52bcdecd97974bf17c43be28a61e4aeb39bfc9b38ff85ef4f9769279499b7ccL455-R587))
* Add segment id parameter to `OrganizationService.findById` method ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-d06de70f55c91875884ab02d75e56977c1d82646913a7d693cbb7cb1a6f9ae79L210-R211))
  * Pass segment id to `OrganizationRepository.findById` method ([link](https://github.com/CrowdDotDev/crowd.dev/pull/1420/files?diff=unified&w=0#diff-d06de70f55c91875884ab02d75e56977c1d82646913a7d693cbb7cb1a6f9ae79L210-R211))

## Checklist ✅
- [ ] ~Add screehshots to the PR description for relevant FE changes~


sausage-todd:
Hey @themarolt,
The segments are designed the way that they are supposed to be transparent to the developer. There is no difference in how they work no matter if they're "enabled" or not. They're basically always enabled.
This is supposed to make engineers life easier when building features, so they don't have to think about more than one scenario.

Is there a benefit in changing the behavior based on the feature flag like in this PR?

backend/src/api/organization/organizationFind.ts
```diff
@@ -20,7 +22,19 @@ import PermissionChecker from '../../services/user/permissionChecker'
 export default async (req, res) => {
   new PermissionChecker(req).validateHas(Permissions.values.organizationRead)
 
-  const payload = await new OrganizationService(req).findById(req.params.id)
+  const segmentId = req.query.segmentId
```
sausage-todd, line 25: Is it some new functionality? 
Originally, segments are coming through the query parameter called `segments`, and there might be many of them.
If you want to get segments coming with the request, you're looking for `req.currentSegments`.
If you want just one, then what you need is `SequelizeRepository.getStrictlySingleActiveSegment(req)`. But keep in mind, this particular request might accept multiple segments.

themarolt: Also because of how OpenSearch is indexed for other query endpoints we can always provide just one segment id and it can be either:
- 1 leaf segment id
- 1 parent segment id
- 1 grandparent segment id
because we have documents for each leaf segment id and then one aggregated for the parent segment id and also one aggregated for grandparent segment id.

This is why this logic is similar - we get only one segmentId as a parameter - matching the logic from all other relevant endpoints.

backend/src/database/repositories/organizationRepository.ts
```diff
@@ -446,81 +446,144 @@ class OrganizationRepository {
...
 
-    const results = await sequelize.query(
+    const replacements: Record<string, unknown> = {
+      id,
+      tenantId: currentTenant.id,
+    }
+
+    let query: string
+
+    if (segmentId) {
```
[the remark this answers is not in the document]
themarolt: This is true - I can make it work with one query and get `segmentId` like you mentioned if none provided.

themarolt:
> Hey @themarolt, The segments are designed the way that they are supposed to be transparent to the developer. There is no difference in how they work no matter if they're "enabled" or not. They're basically always enabled. This is supposed to make engineers life easier when building features, so they don't have to think about more than one scenario.
> 
> Is there a benefit in changing the behavior based on the feature flag like in this PR?



> Hey @themarolt, The segments are designed the way that they are supposed to be transparent to the developer. There is no difference in how they work no matter if they're "enabled" or not. They're basically always enabled. This is supposed to make engineers life easier when building features, so they don't have to think about more than one scenario.
> 
> Is there a benefit in changing the behavior based on the feature flag like in this PR?

We have already this iffed behaviour on member query and organization query endpoint and where we index documents to OpenSearch. This is because we only allow filtering by one segment id (1 leaf or 1 parent or 1 grandparent one) and it's provided as a filter from the frontend. This is because of how OpenSearch works and how we can't really filter on aggregated data - so we pre-build and index aggregated documents for each leaf segment and each parent segment and each grandparent segment. This means that you can't currently filter members/organizations by for example 2 leaf segment ids and 1 parent segment id because there is no such document in the index that has aggregated data for those three segments.

This was more meant to be an error for frontend devs so that they have to provide only one `segmentId` as a filter. On non segmented environments this is not required and as you suggested I can make one query for both cases and get the default segment id to make repo layer segments agnostic. But I would still leave it here on the REST API layer to prevent getting data by default without providing `segmentId` as well. Unless we should make a different endpoint for this case?

sausage-todd:
> This was more meant to be an error for frontend devs so that they have to provide only one `segmentId`as a filter.

Gotcha. Consider using `SequelizeRepository.getStrictlySingleActiveSegment(req)` then. Seems like it does what you need, plus it reuses the functionality correclty.

PR #1848, opened 2023-11-16
Crowd linux

emlimlf:
# Changes proposed ✍️

### What
- Added config for support URL in the integrations page
- Added show/hide config for "Build your own" integrations

### Why
LF users need to be redirected to LF's own Jira support link and the build your own integration is not supported by the LF's instance of crowd.dev


frontend/src/modules/integration/components/integration-list-page.vue
```diff
@@ -99,13 +99,19 @@
   </app-page-wrapper>
 </template>
 
+<script setup>
+import { CrowdIntegrations } from '@/integrations/integrations-config';
+const supportUrl = CrowdIntegrations.getConfig('lfx').supportUrl || 'https://github.com/CrowdDotDev/crowd.dev/issues?q=is%3Aopen+is%3Aissue+label%3A%22Integration+%F0%9F%94%8C%22+';
```
gaspergrom, line 104: I think this is overcomplicated. Please dont use integration config if its not intended to be an actual integration. 
I would just change url in line 13 and thats it.

frontend/src/shared/error/errors.js
```diff
@@ -81,15 +77,15 @@ export default class Errors {
...
-        { duration: 0 },
+        "An error has occurred setting up the integration, please reach out to us via chat.",
+        { duration: 0 }
       );
       return;
     }
 
     Message.error(
-      'Please try again. If the problem remains, reach out to us.',
-      { title: 'Oops, something went wrong' },
```
gaspergrom, line 92: Our linter is forcing to use single quotes so please run `npm run lint:fix`

gaspergrom (changes requested):

frontend/src/shared/error/errors.js
emlimlf: done

frontend/src/modules/integration/components/integration-list-page.vue
emlimlf: done

mlehotskylf:
Can we merge this?

gaspergrom (approved):

PR #1923, opened 2023-12-04
Work experiences when merging

sausage-todd:
Take work-experience merging logic from organizations merging, and reuse it for members merging


epipav (commented):
lgtm :+1:  added 2 nitpick comments

PR #2195, opened 2024-02-15
Simplify services setup

sausage-todd:
I tried to simplify our libs and apps as much as possible. I wanted to:
- make it easier to create new ones
- have more predictability in the ones we already have

### What I improved here so far
1. Moved eslint and prettier to a higher level
    - This includes files:
      - `.eslintrc.cjs`
      - `.prettierignore`
      - `.prettierrc.cjs`
    - and dependencies:
      - `@typescript-eslint/eslint-plugin`
      - `@typescript-eslint/parser`
      - `eslint`
      - `eslint-config-prettier`
      - `eslint-plugin-prettier`
      - `prettier`
    - Less files to copy when creating a new lib/app
    - Having standard centralized configuration
2. Removed typescript aliases for `@crowd/...`
    - These aliases made it possible to use code from a library without actually depending on it
    - Basically, the code from a library was imported through a shortened relative path instead of an npm module
    - We could've as well just removed `package.json` from libraries, and everything would work
    - This explains why when you wanted to use `@crowd/opensearch`, you had to also depend on the actual opensearch library, even though you didn't use it directly
    - Now when you import something from `@crowd/opensearch`, it actually goes through npm module resolution, and brings all the `@crowd/opensearch` dependencies
    - But to make it work, I had to do the next point
3. Replace [`ts-node`](https://github.com/TypeStrong/ts-node) with [`tsx`](https://github.com/privatenumber/tsx)
    - I couldn't make `ts-node` import typescript files from an NPM module
    - Instead, a build step was needed to create javascript files
    - `tsx`, on the other hand, can do that
4. We can now also import code from files other than library's `index.ts`
    - Those `@crowd/...` aliases had a wildcard in place of a library name, which meant we coulnd't use another one to expand everything after `@crowd/opensearch/`
    - Now we can import from any file inside a library with `@crowd/opensearch/src/hello/world.ts`
    - The `src` part looks ugly, but it works and there is no need to stuff everything into `index.ts`
5. Depend on local npm modules through a workspace
    - When writing `"@crowd/common": "file:../../libs/common",` in a `package.json`, `pnpm` didn't realize the dependency was coming from a workspace, so it would copy the dependency into `node_modules`
    - This required a `pnpm install` whenever you change a code of a library
    - It didn't affect us before, because we imported code through shortened relative paths
    - Now with `"@crowd/common": "workspace:*",` `pnpm` actually symlinks the dependencies


services/libs/common/package.json
epipav, line 1: There's logging in `requestThrottler.ts` in common - how would it behave now we're removing logging from common?

services/.eslintrc.cjs
```diff
@@ -17,5 +17,6 @@ module.exports = {
   rules: {
     'prettier/prettier': 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
```
epipav, line 20: This might cause over-usage of any in the future. I think having this on and disabling it if we really should helps with forcing types. What do u think?

services/libs/common/package.json
sausage-todd: It doesn't have any imports from logging:
https://github.com/CrowdDotDev/crowd.dev/blob/c448577c9e7c2da345dae927e3aa7a8f459fa52a/services/libs/common/src/requestThrottler.ts#L1-L4
It uses `any` type for the logger instead:
https://github.com/CrowdDotDev/crowd.dev/blob/c448577c9e7c2da345dae927e3aa7a8f459fa52a/services/libs/common/src/requestThrottler.ts#L8
https://github.com/CrowdDotDev/crowd.dev/blob/c448577c9e7c2da345dae927e3aa7a8f459fa52a/services/libs/common/src/requestThrottler.ts#L18

services/.eslintrc.cjs
sausage-todd: Yeah, true. In that case I'll add exceptions for the places we already have it in

services/libs/common/package.json
epipav: true :+1:

PR #2211, opened 2024-02-22
Unmerge contact

gaspergrom:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


joanagmaia (changes requested):
We should update/fix these:
- As in the designs we should have an unmerge button in the identities drawer (image 1)
- Organizations display in the modal is not correct (image 2)
- If this is going to be merged with main, we should still use "contact" everywhere instead of "contributor". The text updates, should then be done for crowd-linux
- "Change identity" button style is with blue color. When we merge to main, we should still keep the original styles (should be brand color)
- After unmerging identities, I think we should GET the current member being displayed. I think some changes happen immeadiately, so it would be good to reflect them right away.
- "Engagement level" display is not correct in modal. Should be as unknown for both (image 3)
- I've waited for a couple of minutes but I never got the WS message with the confirmation. Were you able to test this?

Screenshot 2024-02-22 at 13 09 59

Screenshot 2024-02-22 at 13 10 56

Screenshot 2024-02-22 at 13 13 12

joanagmaia (approved):

PR #2224, opened 2024-02-26
Improvement/lf contributor unmerge

gaspergrom:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


frontend/src/modules/member/member-service.js
```diff
@@ -229,6 +229,31 @@ export class MemberService {
     return response.data;
   }
 
+  static async unmerge(memberId, preview) {
+    const tenantId = AuthCurrentTenant.get();
+
+    const response = await authAxios.post(
```
joanagmaia, line 235: Can you sync with Anil about which segments we should pass in the request?
Since we are not explicitly passing any, by default it will send all sub projects ids of the selected project group. Not sure that's what we want

epipav: we're not using segments on merge/unmerge bcs we already know what segments a member belongs to. If a user can see the member in current segment, they should be able to unmerge it, and we deduce the unmerged members segments from its activities that were moved

joanagmaia: @epipav makes sense. If it's an extraction and we don't know the segments we will create the member in the same segments as the original one?

epipav: yes, we sync the member to opensearch to segments that the primary member is in. We also set the `member.manuallyCreated` to true

joanagmaia: Sounds good

joanagmaia (approved):

PR #2410, opened 2024-05-16
use bulkRemove for OpenSearch index cleanup

yeganathan18:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


services/libs/opensearch/src/service/member.sync.service.ts
```diff
@@ -135,9 +135,7 @@ export class MemberSyncService {
 
       if (toRemove.length > 0) {
         this.log.warn({ tenantId, toRemove }, 'Removing members from index!')
-        for (const id of toRemove) {
-          await this.openSearchService.removeFromIndex(id, OpenSearchIndex.MEMBERS)
-        }
+        await this.openSearchService.bulkRemoveFromIndex(toRemove, OpenSearchIndex.MEMBERS)
```
themarolt, line 138: maybe go even further and gather ids in toRemove until you reach the maximum allowed for bulk remove?

themarolt (changes requested):

services/libs/opensearch/src/service/member.sync.service.ts
yeganathan18: Good point! There's no limit for bulk removal in OpenSearch; we can do as much as we want, but that also affects the time taken to do the operation on the passed IDs.

I thought this should be safe and quick 🤔

themarolt: What I mean was:
currently we do cleanup by `tenantId` like this:
```
while (nextPage) {
  memberIds = fetchFromOpensearch
  memberIdsNotInDb = filterThroughDb(memberIds)
  delete memberIdsNotInDb from OpenSearch
}  
```

what if we do smt like this:
```
memberIdsNotInDb = []
while (nextPage) {
  memberIds = fetchFromOpensearch
  memberIdsNotInDb.push(filterThroughDb(memberIds))
  if (memberIdsNotInDb.length > SOME_THRESHOLD) {
      delete memberIdsNotInDb from OpenSearch
      memberIdsNotInDb = []
   }
} 

if (memberIdsNotInDb.length > 0) {
 delete memberIdsNoInDb from OpenSearch
} 
```

this is to eliminate too many OpenSearch delete/bulkDelete calls - maybe test it out a bit to see if it pays out on the long run for bigger tenants...

yeganathan18: Ah gotcha, maybe we can have `SOME_THRESHOLD` as 300.

yeganathan18:
@themarolt Since we already cleaned up the organizations in OpenSearch, we couldn't test this properly as it's already pretty fast with fewer items to remove 🙂

But it's much better than before!

themarolt (approved):

PR #2458, opened 2024-06-10
Enhancement/org identity types lfx 1129

themarolt:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


themarolt:
@epipav Frontend changes were already reviewed in https://github.com/CrowdDotDev/crowd.dev/pull/2453.

services/libs/data-access-layer/src/old/apps/data_sink_worker/repo/organization.repo.ts
```diff
@@ -130,21 +51,20 @@ export class OrganizationRepository extends RepositoryBase<OrganizationRepositor
               select oi."organizationId"
               from "organizationIdentities" oi
               where 
-                    oi.platform = $(platform)
-                    and oi.name ilike $(name)
+                    oi."tenantId" = $(tenantId)
+                    and oi.platform = $(platform)
+                    and oi.value ilike $(value)
```
epipav, line 56: let's use `lower(oi.value) = lower($(value))` match here instead (like we do in members) it's performing much better with an index on lower(value)

services/libs/integrations/src/integrations/premium/hubspot/field-mapper/organizationFieldMapper.ts
```diff
@@ -269,13 +234,16 @@ export class HubspotOrganizationFieldMapper extends HubspotFieldMapper {
       )
     }
 
+    // TODO uros check if this is verified or not with anil
```
epipav, line 237: we can assume it's verified if it's coming from hubspot i guess

services/libs/integrations/src/integrations/premium/hubspot/field-mapper/memberFieldMapper.ts
```diff
@@ -204,12 +205,16 @@ export class HubspotMemberFieldMapper extends HubspotFieldMapper {
               })
             }
           } else if (crowdKey === 'organizationName') {
+            // TODO uros check if this is verified or not with anil
```
epipav, line 208: should be verified

services/libs/integrations/src/integrations/premium/hubspot/api/batchCreateOrganizations.ts
```diff
@@ -109,45 +116,46 @@ export const batchCreateOrganizations = async (
...
+      //     )
+      //     return await batchCreateOrganizations(
+      //       nangoId,
+      //       createOrganizations,
+      //       organizationMapper,
+      //       ctx,
+      //       throttler,
+      //     )
+      //   }
+      // }
```
epipav, line 158: Idea here was if we tried creating an existing organization HubSpot would return a conflict error - We parse the conflicting id for the batch create request and update these conflicting ones instead (removing the conflicting id from the batch create and retrying)
Since HubSpot integration is not used anymore, I'm not sure how relevant it's to try to maintain it with the changing code tbh

epipav (commented):
Org identities make much more sense now :muscle: Added comments & few nitpicks

services/apps/premium/members_enrichment_worker/src/activities/normalizeEnrichedMember.ts
```diff
@@ -117,41 +121,51 @@ export async function updateOrganizations(input: EnrichingMember): Promise<strin
...
-
-          const organizationIdentities: IOrganizationIdentity[] = [
+          const identities: IOrganizationIdentity[] = [
             {
-              name: workExperience.company,
-              platform: PlatformType.ENRICHMENT,
+              platform: PlatformType.LINKEDIN,
+              value: workExperience.companyUrl,
+              type: OrganizationIdentityType.PRIMARY_DOMAIN,
+              verified: true,
```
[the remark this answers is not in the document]
themarolt: Not for domains - just for usernames.

services/libs/integrations/src/integrations/premium/hubspot/field-mapper/organizationFieldMapper.ts
themarolt: It actually doesn't matter - I checked with Joan and LF is not gonna be using HubSpot and it's probably gonna get removed (the code for integration) so at this point we don't have to touch it.

backend/src/services/organizationService.ts
```diff
@@ -284,20 +300,31 @@ export default class OrganizationService extends LoggerBase {
...
+          const identity = organization.identities.find(
+            (i) => i.type === OrganizationIdentityType.PRIMARY_DOMAIN && i.value === domain,
+          )
+          await OrganizationRepository.removeIdentitiesFromOrganization(
+            organizationId,
+            [identity],
+            repoOptions,
+          )
+        }
       }
```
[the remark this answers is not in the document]
themarolt: The problem is that in the db we probably don't have the verified flag yet and there is no easy way to add it...
So the unmerge backup won't have the correct structure - that's why in the repo layer I kinda try to recreate the correct structure (with value, type instead of name) but the verified flag is not set...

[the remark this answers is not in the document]
themarolt: That's why I only check verified flag for identities coming from org that came from db.

PR #2495, opened 2024-06-27
Return avg contributor engagement in organization profile

yeganathan18:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/backend/src/database/migrations/U1719390926__orgAvgContributorEngagement.sql
+++ b/backend/src/database/migrations/U1719390926__orgAvgContributorEngagement.sql
@@ -1,1 +1,0 @@
-ALTER TABLE "organizationSegmentsAgg" DROP COLUMN "avgContributorEngagement";
--- a/backend/src/database/migrations/U1719390926__orgAvgContributorEngagement.sql
+++ b/backend/src/database/migrations/U1719390926__orgAvgContributorEngagement.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "organizationSegmentsAgg" DROP COLUMN "avgContributorEngagement";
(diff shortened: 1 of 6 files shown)


backend/src/database/migrations/U1719390926__orgAvgContributorEngagement.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "organizationSegmentsAgg" DROP COLUMN "avgContributorEngagement";
```
sausage-todd, line 1: We don't use rollbacks, so no need in these

backend/src/database/migrations/V1719390926__orgAvgContributorEngagement.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE "organizationSegmentsAgg" ADD COLUMN "avgContributorEngagement" INTEGER DEFAULT '-1'::INTEGER;
```
sausage-todd, line 1: Let's do it this way:
1. Have the field non-nullable like every other aggregated field in this table
2. So, no arbitrary default values, which if they come up at some point in the UI, would look confusing
3. So, drop the default value here
4. In the end of the migration, after populating values, make the field non-nullable

sausage-todd (commented):
Looks good apart from the nullable field

backend/src/database/migrations/V1719390926__orgAvgContributorEngagement.sql
yeganathan18: Quick question: If we make it non-nullable, shouldn’t we handle it in the code if the aggregation is null? 

For example, if the `avgContributorEngagement` is null for an organization and segment, can we use COALESCE and fallback to 0?

sausage-todd: If there are activities, aggregation can’t be null. If there are no activities, the org/segment will be excluded from the query.

PR #2502, opened 2024-07-02
Recompute organization aggs for new incoming activities

yeganathan18:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/services/apps/data_sink_worker/src/service/activity.service.ts
+++ b/services/apps/data_sink_worker/src/service/activity.service.ts
@@ -757,6 +757,9 @@ export default class ActivityService extends LoggerBase {
                 false,
               )
 
+              // cache the organization for aggregate recomputation
+              await this.redisClient.sAdd('computeOrgAgg', organizationId)
+
               activityId = dbActivity.id
             }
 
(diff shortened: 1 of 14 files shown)


services/apps/data_sink_worker/src/service/activity.service.ts
```diff
@@ -951,6 +952,10 @@ export default class ActivityService extends LoggerBase {
       if (activityId) {
         await this.searchSyncWorkerEmitter.triggerActivitySync(tenantId, activityId, onboarding)
       }
+      if (organizationId) {
+        this.log.info({ organizationId }, 'Caching organization for aggregate recomputation!')
```
epipav, line 956: Let's make sure to reduce the log level here before deploying

epipav (changes requested):
- pnpm lock file version looks like it changed. Did you run pnpm install on a lower version of node? (it should be v20)
- lets execute the child workflows 10 by 10, waiting each 10 instead of running it for all to keep db load low. what do u think?	
- NIT: can we use a more descriptive key? something like organizationIdsForAggComputation or similar

epipav (approved):

PR #2515, opened 2024-07-08
More convenient segments with: `parentId`, `grandparentId` and `type`

epipav:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
+++ b/backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
@@ -9,7 +9,6 @@ ALTER TABLE "segments" ADD COLUMN "type" TEXT GENERATED ALWAYS AS (
     ) STORED;
 
 
-BEGIN;
 DO
 $$
     DECLARE
--- a/backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
+++ b/backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
@@ -1,0 +1,56 @@
+ALTER TABLE "segments" ADD COLUMN IF NOT EXISTS "parentId" uuid;
+ALTER TABLE "segments" ADD COLUMN IF NOT EXISTS "grandparentId" uuid;
+ALTER TABLE "segments" ADD COLUMN "type" TEXT GENERATED ALWAYS AS (
+        CASE
+            WHEN "parentSlug" IS NULL AND "grandparentSlug" IS NULL THEN 'projectGroup'
+            WHEN "grandparentSlug" IS NULL AND "parentSlug" IS NOT NULL THEN 'project'
+            ELSE 'subproject' 
+        END
+    ) STORED;
+
+
+BEGIN;
+DO
+$$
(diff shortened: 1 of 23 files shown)


backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
```diff
@@ -0,0 +1,56 @@
...
+ALTER TABLE "segments" ADD COLUMN "type" TEXT GENERATED ALWAYS AS (
+        CASE
+            WHEN "parentSlug" IS NULL AND "grandparentSlug" IS NULL THEN 'projectGroup'
+            WHEN "grandparentSlug" IS NULL AND "parentSlug" IS NOT NULL THEN 'project'
+            ELSE 'subproject' 
+        END
+    ) STORED;
+
+
+BEGIN;
```
sausage-todd, line 12: Anonymous blocks are executed in a transaction anyway, so no need

backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
```diff
@@ -0,0 +1,56 @@
...
+                ELSIF segment."type" = 'subproject' THEN
+                    SELECT * INTO parent
+                    FROM segments
+                    WHERE "slug" = segment."parentSlug"
+                    AND "type" = 'project';
+
+                    SELECT * INTO grandparent
+                    FROM segments
+                    WHERE "slug" = segment."grandparentSlug"
+                    AND "type" = 'projectGroup';
```
sausage-todd, line 42: Have you tried running this on production? At least without the updates, just with something like
```
RAISE NOTICE 'updating parent id to %', parent.id;
```

I'm afraid this kind of filtering won't be enough. Since we don't have a unique index on `slug`, there might be many different segments on the same level with the same slug. And if we don't check for parent slugs too, we might get inconsistencies.

I think if you just run it on production without updates, and if I'm right about inconsistencies, then queries like
```
                    SELECT * INTO parent
                    FROM segments
                    WHERE "slug" = segment."parentSlug"
                    AND "type" = 'project';
```
will fail because they'll try to select multiple rows into one variable.

sausage-todd (approved):
Looks good

backend/src/database/migrations/V1720446425__segmentsWithParentIdsAndType.sql
epipav: we have these indexes 
```
CREATE UNIQUE INDEX "segments_slug_pSlug_gpSlugNull_tenantId_idx" ON "segments" ("slug", "parentSlug", "tenantId") WHERE "grandparentSlug" IS NULL;
CREATE UNIQUE INDEX "segments_slug_pSlugNull_gpSlugNull_tenantId_idx" ON "segments" ("slug", "tenantId") WHERE "grandparentSlug" IS NULL AND "parentSlug" IS NULL;

```
that guarantees slug uniqueness in project group and project levels

I'll try it on prod to see if there are any issues :+1:

PR #2519, opened 2024-07-09
Member aggregates api

sausage-todd:


backend/src/database/repositories/memberRepository.ts
```diff
@@ -1273,20 +1297,19 @@ class MemberRepository {
       // we load data for a specific segment (can be leaf, parent or grand parent id)
 
       const dataFromOpensearch = (
```
epipav, line 1299: NIT: let's use a different variable name here to prevent confusion

backend/src/database/repositories/memberRepository.ts
```diff
@@ -2313,7 +2170,237 @@ class MemberRepository {
...
+        withAggregates
+          ? `LEFT JOIN "memberSegmentsAgg" msa ON msa."memberId" = m.id AND msa."segmentId" = $(segmentId)`
+          : ''
+      }
+      ${
+        include.memberOrganizations
+          ? `LEFT JOIN "memberOrganizations" mo ON mo."memberId" = m.id`
+          : ''
+      }
+      WHERE 1=1
```
epipav, line 2301: do we need this? there are already some default filters

backend/src/services/memberService.ts
```diff
@@ -862,22 +868,41 @@ export default class MemberService extends LoggerBase {
...
-      )
+      const member = {
+        ...memberById,
+        memberOrganizations,
+        identities,
+        affiliations,
+        tags: tags.map((t) => ({ id: t.tagId })),
+        notes: notes.map((n) => ({ id: n.noteId })),
+        tasks: tasks.map((t) => ({ id: t.taskId })),
+        emails: [], // I don't think we can unmerge these after moving to them attributes
```
epipav, line 901: I think we can remove this field from `IMemberUnmergePreviewResult`

epipav (approved):
Added few nitpicks, lgtm

backend/src/database/repositories/memberRepository.ts
sausage-todd: True, we don't

backend/src/services/memberService.ts
sausage-todd: Ok

PR #2531, opened 2024-07-19
Person create

gaspergrom:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


frontend/src/modules/contributor/components/edit/contributor-add.vue
```diff
@@ -0,0 +1,264 @@
...
+        },
+        query: {
+          projectGroup: selectedProjectGroup.value?.id,
+        },
+      });
+      isModalOpen.value = false;
+    })
+    .catch(() => {
+      Message.error('There was an error creating a person');
+    })
```
joanagmaia, line 253: There is a very specific error that backend shows: 400 "Member with the same platform and username already exists in this tenant".

where we shouldn't simply show "there was an error creating a person".

Can we show a better message for when this happens? Suggestion: "We couldn't add a new person to your community, because one of the identities already exist in another profile."
Also if you check the code, we already have backend returning 409 instead of 400 when this happens. So perhaps you can update the backend error code so that frontend can handle the message display a bit better.

And as a fallback message: "There was an error adding a new person to your community."

frontend/src/modules/contributor/components/edit/contributor-add.vue
```diff
@@ -0,0 +1,264 @@
...
+  }));
+
+interface ContributorAddForm {
+  subproject: string;
+  name: string;
+  email: string[];
+  identities: ContributorAddIdentity[];
+}
+
+const form = reactive<ContributorAddForm>({
```
joanagmaia, line 181: Initially when you open the drawer you have everything disabled until the user selects the sub-project. If you select it, everything becomes enabled.
However, if you close the modal and open it after, everything is enabled but sub-project is not selected.

Perhaps when the user closes the modal, everything should be cleaned and disabled again

joanagmaia (approved):

PR #2545, opened 2024-07-25
Integration visibility

garrrikkotua:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/frontend/src/integrations/confluence/components/confluence-connect.vue
+++ b/frontend/src/integrations/confluence/components/confluence-connect.vue
@@ -1,5 +1,5 @@
 <template>
-  <slot :connect="connect" :settings="settings" :has-settings="hasSettings" />
+  <slot :connect="connect" :settings="settings" :has-settings="hasSettings" :settings-component="ConfluenceSettings" />
   <app-confluence-connect-drawer
     v-model="drawerVisible"
     :integration="integration"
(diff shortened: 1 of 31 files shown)


joanagmaia (commented):
From QA:
- For Github, can you also add a title to the popover? So basically “GitHub repositories”
- For Github, for the repo, can you make it linkable? Opening the repo in another tab
- For Git, trim the remote url, and only show the repo name. Also make it clickable
- For HackerNews, shouldn’t it be posts instead of urls?
- For Slack, when all channels are connected we shouldn’t show a popover. Instead of 0 channels we should show “All channels” (I’ll run this one by Nuno as well)
- For Stack Overflow (and all other platforms that have more than 1 section in the popover), we should not show the section if it’s empty. E.g. for Stack Overflow we show “3 tags . 0 keywords”. Then on the popover below Stack Overflow keywords we show nothing. In this scenarios I would say not to show the value, so it would only be “3 tags” and the popover we wouldn’t show the keywords section

Couple of things that can be improved:
- For icons instead of using <div>, we usually use <i>
- For font-size, instead of using text-[13px] we usually use e.g. text-small. You can check the available values in `tailwind.config.js`
- 
- Why are we adding `webkit-scrollbar` styling to all settings components? Do we even need to style the scrollbar?
- For all integrations that have repository links, git, github, and gerrit, can we actually make them linkable. So for instance, for the list of git repositories, if the user clicks on a repository in the popover it's a link that will open in a new tab.
- Instead of doing, e.g.
```
{{ groups.length }}
{{ groups.length !== 1 ? "groups" : "group" }}
```
You could do
```
pluralize('group', groups.length, true)
```
This will render as: 2 groups or 1 group

joanagmaia (approved):

PR #2568, opened 2024-08-07
Improve performance of searching members by name

sausage-todd:
Split filtering on search value to avoid double seq-scan on `members` table


sausage-todd:
Merged into https://github.com/CrowdDotDev/crowd.dev/pull/2566

PR #2572, opened 2024-08-09
Update lfx membership script

joanagmaia:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/backend/src/bin/scripts/import-lfx-memberships.ts
+++ b/backend/src/bin/scripts/import-lfx-memberships.ts
@@ -119,38 +119,44 @@ if (parameters.help || !parameters.file || !parameters.tenantId) {
 
     for (let i = 0; i < records.length; i++) {
       const record = records[i]
-      record['Domain Alias'] = parseDomains(record['Domain Alias'])
-
       const orgName = record['Account Name']
 
-      const segment = await findProjectGroupByName(qx, {
-        tenantId,
-        name: record['Project'],
-      })
-      const orgId = await findOrgId(qx, tenantId, record)
-      const row = {
-        tenantId: parameters.tenantId,
-        organizationId: orgId,
-        segmentId: segment?.id,
-        accountName: orgName,
-        parentAccount: record['Parent Account'],
-        project: record['Project'],
-        productName: record['Product Name'],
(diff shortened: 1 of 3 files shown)


services/libs/data-access-layer/src/organizations/organizations.ts
```diff
@@ -69,22 +69,22 @@ export async function findOrgIdByDisplayName(
...
-        AND "deletedAt" IS NULL
+        AND "type" IN ('${OrganizationIdentityType.PRIMARY_DOMAIN}', '${OrganizationIdentityType.ALTERNATIVE_DOMAIN}')
       LIMIT 1;
     `,
     {
-      websites,
+      domains,
       tenantId,
     },
   )
```
sausage-todd, line 90: Move this to another file called `identities.ts` in the same directory, as it's working with a different table now

backend/src/bin/scripts/import-lfx-memberships.ts
```diff
@@ -101,6 +101,11 @@ if (parameters.help || !parameters.file || !parameters.tenantId) {
     const qx = SequelizeRepository.getQueryExecutor({
       database: prodDb,
     } as IRepositoryOptions)
+
+    await qx.result(`TRUNCATE "lfxMemberships"`)
```
sausage-todd, line 105: This can be dangerous if there are other tables *depending* on this one, and the will be truncated too.
Let's change it to `DELETE FROM "lfxMemberships"`, so if we actually have some dependencies, we know about it, because this query starts to fail.

sausage-todd (approved):
Also, I remember in the LFX memberships data there was an entry for an individual contributor (or anonymous something, don't remember exactly). Not even sure why they have it in the data, but it would display that individual contributor org profile as an LFX member. We then deleted manually this entry from the table.

It might make sense to do something about this entry when re-importing the data again:
- either filter it before insertion;
- or delete after.

Otherwise, looks great

joanagmaia:
@sausage-todd thank you so much Misha!
And that's a very valid point with "Individual - No account", I'm now excluding it.
Updated the PR

PR #2573, opened 2024-08-13
Organization add

gaspergrom:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/frontend/src/modules/organization/components/edit/organization-add.vue
+++ b/frontend/src/modules/organization/components/edit/organization-add.vue
@@ -1,0 +1,261 @@
+<template>
+  <lf-modal v-model="isModalOpen">
+    <!-- Header -->
+    <section class="py-4 pr-4 pl-6 flex justify-between items-center">
+      <h5>Add organization</h5>
+      <lf-button type="secondary-ghost-light" :icon-only="true" @click="isModalOpen = false">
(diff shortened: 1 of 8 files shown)


frontend/src/modules/organization/components/edit/organization-add.vue
```diff
@@ -0,0 +1,261 @@
...
+        query: {
+          projectGroup: selectedProjectGroup.value?.id,
+        },
+      });
+      isModalOpen.value = false;
+      resetForm();
+    })
+    .catch((error: any) => {
+      if (error.response.status === 409) {
+        Message.error('Organization was not created because the identity already exists in another profile');
```
joanagmaia, line 246: Can you update this and the `contributor-identity-add`, so that below the toast message we have a button to View the profile? Backend returns the id right?

joanagmaia: Actually I just tried and I believe they already return the existing organization in the payload. So this error message won't even exist.
Do you think there's a way to distinguish if the organization already existed or not? Because that way we could show an informative message mentioning that an organization with the same identity already existed

joanagmaia (approved):

PR #2577, opened 2024-08-14
adjust the findById endpoint to show an org across all segments

yeganathan18:
# Changes proposed ✍️

The current behavior of the organization.findById endpoint is that if no segmentId is passed, it returns data (aggregates) across all segments.


sausage-todd (approved):
👍

PR #2694, opened 2024-11-21
Fix merging of big orgs

sausage-todd:
This PR fixes merging for orgs that have a lot of members.

The actual problem with them is this:
1. `finishOrgMerging` temporal workflow calls `searchSyncApi.syncOrgMembers(orgId)`
3. `search-sync-api` finds all org members, and sync them one by one
4. Some big orgs have more than 1000 members. Calculating aggregates for one member with questdb is around 20 seconds. That's about 5 hours.
5. `finishOrgMerging` temporal workflow has a limit of 15 minutes for activity execution
6. Temporal activity fails, restarts, we repeat from point 1

How it's fixed:
1. We don't sync members which were already synced in the previous attempt by remembering the time when the sync was requested and `updatedAt` of rows in `memberSegmentsAgg`
2. We increase activity execution timeout in `finishOrgMerging` temporal workflow from 15 minutes to 1 hour
3. When syncing all org members, we don't pre-filter members by taking only those who have activities. This is not needed, because calculating aggregates takes care of that

## Release Notes

- **New Features**
  - Added a `createdAt` timestamp column to track record creation in the member segments.
  - Introduced a `syncFrom` parameter for improved synchronization control in organization member syncing.
  - Enhanced synchronization methods to respect the last sync date, optimizing data handling.
  - Increased timeout duration for merging and unmerging activities to allow longer execution times.

- **Bug Fixes**
  - Adjusted timeout settings for merging and unmerging activities to allow longer execution times.

- **Documentation**
  - Updated method signatures to reflect new parameters and functionality for better clarity.


epipav (approved):
looks good - ts build is failing tho

PR #2740, opened 2025-01-03
Remove unused features, configs, and sample data

skwowet:
Removed following items in backend:
 - quickstart guide
 - notes and tasks logic
 - plans & stripe related code
 - api docs & openapi
 - feature-flags + unleash
 - sample data
 - emails-worker (responsible for weekly emails)
 - unused user and auth endpoints


scripts/deploy/kubernetes/.backend.env
```diff
@@ -34,12 +34,6 @@ CROWD_DB_USERNAME=postgres
 CROWD_DB_PASSWORD=example
 CROWD_DB_DATABASE=crowd-web
 
-CROWD_UNLEASH_DB_HOST=unleash-db-svc
-CROWD_UNLEASH_DB_PORT=5432
```
epipav, line 38: double check with @themarolt but I think we don't use /deploy/kubernetes folder anymore we can get rid of it. These files live in crowd-kube now

services/apps/cache_worker/src/workflows/spawnDashboardCacheRefreshForAllTenants.ts
```diff
@@ -17,112 +16,84 @@ const activity = proxyActivities<typeof activities>({ startToCloseTimeout: '1 mi
...
   const tenants = await activity.getAllTenants()
-  const segmentsEnabled = await activity.isSegmentsEnabled()
   const info = workflowInfo()
 
-  if (segmentsEnabled) {
-    // we should spawn refreshDashboardCache for each tenant-segment couples
+  const SEGMENT_PAGE_SIZE = 250
 
-    const SEGMENT_PAGE_SIZE = 250
+  for (const tenant of tenants) {
```
epipav, line 23: can u run `npm run format` for this service?

epipav (commented):
- We can also get rid of quickstart
- should we remove fields that are not used anymore? like `tenant.plan tenant.stripeSubsId` `tenant.planSubsEndsAt` etc.. from db?
- Also we should remove tables that are not used anymore (`tasks`, `notes` `activityTasks` etc)
- I suggest we remove all existing tests too, they're just deadweight

services/apps/cache_worker/src/workflows/spawnDashboardCacheRefreshForAllTenants.ts
skwowet: done

scripts/deploy/kubernetes/.backend.env
skwowet: Done ✅

backend/config/custom-environment-variables.json
```diff
@@ -169,22 +150,6 @@
   "slackAlerting": {
     "url": "CROWD_SLACK_ALERTING_URL"
   },
-  "sampleData": {
-    "tenantId": "CROWD_SAMPLE_DATA_TENANT_ID"
-  },
-  "unleash": {
```
themarolt, line 175: We still have unleash deployed on production - AFAIK we will still need it for feature flags for testing.

skwowet: Got green signal from Joana for removing it, since LF uses a different vendor. If we need feature flags in the future, we can probably build them from scratch ✌🏻

PR #2846, opened 2025-02-15
Programmatically download Kafka Connect HTTP

borfast:
# Changes proposed ✍️

The downloaded .jar file for Kafka Connect HTTP is almost 100 MB. Instead of adding a huge file to our repository, we could automate downloading it if it doesn't exist, when running the CLI `scaffold up` command.

In order to really do this, we should also amend the commit that introduced the files and rebase the branch, otherwise they will always stay in the repository history and there will be no space savings.

### Why

Having big files in a git repository is usually a bad practice because it makes all work on the repository slower, from cloning it (which can be particularly painful in CI environments, where it is usually always cloned from scratch), to running diffs, and other issues. It's also less wasted storage space in all our machines and whatever servers the repository files will end up.

## Checklist ✅


epipav (changes requested):
nice! can u also add the files we download to .gitignore

borfast:
> nice! can u also add the files we download to .gitignore

Done. It shouldn't be necessary because it is removed at the end of the `download_kafka_connect_http` function, but it also doesn't hurt.

epipav:
> > nice! can u also add the files we download to .gitignore
> 
> Done. It shouldn't be necessary because it is removed at the end of the `download_kafka_connect_http` function, but it also doesn't hurt.

yes you're deleting the zip file, but the unzipped stuff are still tracked by git and needs to be ignored as well right?

borfast:
Oh, you're absolutely right, I didn't think of those! :facepalm: 
Done. Thanks for catching that! :+1:

epipav (approved):

PR #2943, opened 2025-03-18
Add `query-stream` config and error propagation 

skwowet:
# Changes proposed ✍️

- Explicitly added `batchSize` and `highWaterMark` options to pg-query-stream for performance.
- Fix stream error handling to propagate failures to Temporal workflows, enabling proper retries.


skwowet:
> What do they do in short? Link to docs?

The stream configuration optimizes memory usage and performance:
- `batchSize`: number of rows fetched from db per batch
- `highWaterMark`: controls the maximum number of rows that will be buffered in memory by the Node.js stream

Link to docs: https://github.com/brianc/node-postgres/tree/master/packages/pg-query-stream

sausage-todd (approved):

PR #2953, opened 2025-03-25
Trigger merge suggestions on profile updates

skwowet:
# Changes proposed ✍️

Previously, merge suggestions were only generated for newly created members and organizations. Now, they're generated when existing profiles are modified as well.

Here's what changed:
- generateMergeSuggestion now uses `updatedAt` instead of `createdAt` for members and organizations.
- added `updatedAt` timestamp update when member identities are modified.
- `merge-suggestion-worker` now processes all profile modifications during scheduled runs.

This fixes gaps where profile updates weren’t triggering new merge suggestions.


services/libs/data-access-layer/src/members/others.ts
```diff
@@ -78,3 +78,7 @@ export async function updateMemberReach(
     },
   )
 }
+
+export async function setMemberUpdatedAt(qx: QueryExecutor, memberId: string): Promise<void> {
```
epipav, line 82: can we instead call this `touchMemberUpdatedAt` since we're not giving a date as input and set to that value, but instead setting it to `now()`

epipav (commented):
lgtm, added one nitpick :+1:

PR #2969, opened 2025-04-02
Category group management

gaspergrom:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


backend/src/services/categoryService.ts
```diff
@@ -0,0 +1,266 @@
...
+   * Creates a new category group with optional associated categories.
+   *
+   * @param {ICreateCategoryGroupWithCategories} categoryGroup - An object representing the category group to be created, which may include associated categories.
+   * @return {Promise<boolean>} A promise that resolves to true if the category group is successfully created.
+   */
+  async createCategoryGroup(categoryGroup: ICreateCategoryGroupWithCategories) {
+    return SequelizeRepository.withTx(this.options, async (tx) => {
+      const qx = SequelizeRepository.getQueryExecutor(this.options, tx)
+
+      let slug = slugify(categoryGroup.name, {
```
themarolt, line 50: no need for `slugify` please check `generateSlug` function in `createConversations.ts` - we have a common method to clean a string.

services/libs/data-access-layer/src/categories/index.ts
```diff
@@ -0,0 +1,501 @@
...
+  filters: ICategoryGroupsFilters,
+): Promise<ICategoryGroup[]> {
+  return qx.select(
+    `
+            SELECT id, name, type
+            FROM "categoryGroups"
+            WHERE ($(type) IS NULL OR type = $(type))
+              AND ($(query) = '' OR name ILIKE CONCAT('%', $(query), '%'))
+            ORDER BY name
+            LIMIT $(limit) OFFSET $(offset)
```
skwowet, line 126: Can we optimize this query like this?

```sql
SELECT id, name, type
FROM "categoryGroups"
WHERE COALESCE($(type), type) = type
  AND ($(query) = '' OR name ILIKE $(searchPattern))
ORDER BY name
LIMIT $(limit) OFFSET $(offset)
```

Key improvements:
1. Replaced `($(type) IS NULL OR type = $(type))` with `COALESCE($(type), type) = type`
   - This is more efficient because it avoids the OR condition which can prevent pgsql from using indexes effectively
   - COALESCE is a single operation vs two conditions with OR

2. Changed `CONCAT('%', $(query), '%')` to use a parameter `$(searchPattern)`
   - We should pass `searchPattern` as `%${query}%` from the code
   - This moves the string concatenation out of the sql layer where it would run for every row
   - Postgres can better optimize the query when it receives the complete pattern

These changes will be especially noticeable when dealing with larger datasets or under high load.

Let me know what you think!

services/libs/data-access-layer/src/categories/index.ts
```diff
@@ -0,0 +1,501 @@
...
+export async function listCategoryGroupsCount(
+  qx: QueryExecutor,
+  filters: ICategoryGroupsFilters,
+): Promise<number> {
+  const result = await qx.selectOne(
+    `
+            SELECT COUNT(*) as count
+            FROM "categoryGroups"
+            WHERE ($(type) IS NULL OR type = $(type))
+              AND ($(query) = '' OR name ILIKE CONCAT('%', $(query), '%'))
```
skwowet, line 177: Same optimization here

services/libs/data-access-layer/src/categories/index.ts
```diff
@@ -0,0 +1,501 @@
...
+    name: string
+    categoryGroupId: string
+    categoryGroupName: string
+  }[]
+> {
+  return qx.select(
+    `          SELECT c.id, c.name, cg.id as "categoryGroupId", cg.name as "categoryGroupName"
+                   FROM "categories" c
+                            JOIN "categoryGroups" cg ON c."categoryGroupId" = cg.id
+                   WHERE c.name ILIKE CONCAT('%', $(query), '%')
```
skwowet, line 469: Same here

skwowet (approved):

Sameh16 (approved):
FE-LGTM

PR #2970, opened 2025-04-02
fix: resolve project to sub-project segments in member and org queries

skwowet:
# Changes proposed ✍️

Modified member queries to automatically resolve project/group segment Ids to their relevant sub-project segment Ids, ensuring members are returned regardless of which level of segment ID is provided.

### DIFF
--- a/backend/src/database/repositories/organizationRepository.ts
+++ b/backend/src/database/repositories/organizationRepository.ts
@@ -38,6 +38,7 @@ import { findAttribute } from '@crowd/data-access-layer/src/organizations/attrib
 import { optionsQx } from '@crowd/data-access-layer/src/queryExecutor'
 import {
   findSegmentById,
+  getSegmentSubprojectIds,
   isSegmentProject,
   isSegmentProjectGroup,
 } from '@crowd/data-access-layer/src/segments'
@@ -1566,8 +1567,9 @@ class OrganizationRepository {
       }
     }
 
+    let segments = []
     if (segmentId) {
-      const segment = (await findSegmentById(optionsQx(options), segmentId)) as any
+      const segment = (await findSegmentById(optionsQx(options), segmentId))
 
       if (segment === null) {
         options.log.info('No segment found for organization')
@@ -1579,13 +1581,13 @@ class OrganizationRepository {
         }
       }
 
-      segmentId = segment.id
+      segments = getSegmentSubprojectIds(segment)
     }
 
     const params = {
       limit,
       offset,
-      segmentId,
+      segments,
       tenantId: options.currentTenant.id,
     }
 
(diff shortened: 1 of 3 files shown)


backend/src/database/repositories/organizationRepository.ts
```diff
@@ -1579,13 +1581,13 @@ class OrganizationRepository {
         }
       }
 
-      segmentId = segment.id
+      segments = getSegmentSubprojectIds(segment)
```
epipav, line 1584: NIT: can we call this `segmentIds` instead? It's ambiguous if this has ids or the full object this way

backend/src/database/repositories/organizationRepository.ts
```diff
@@ -1579,13 +1581,13 @@ class OrganizationRepository {
...
     }
 
     const params = {
       limit,
       offset,
-      segmentId,
+      segments,
```
epipav, line 1590: same

services/libs/data-access-layer/src/members/base.ts
```diff
@@ -129,9 +134,9 @@ export async function queryMembersAdvanced(
   }
 
   const withAggregates = !!segmentId
-  let segment
+  let segments = []
```
epipav, line 137: same

backend/src/database/repositories/organizationRepository.ts
```diff
@@ -1617,12 +1619,8 @@ class OrganizationRepository {
       FROM organizations o
       ${
         withAggregates
-          ? ` INNER JOIN "organizationSegmentsAgg" osa ON osa."organizationId" = o.id AND ${
-              segmentId ? `osa."segmentId" = $(segmentId)` : `osa."segmentId" IS NULL`
-            }`
-          : ` LEFT JOIN "organizationSegmentsAgg" osa ON osa."organizationId" = o.id AND ${
-              segmentId ? `osa."segmentId" = $(segmentId)` : `osa."segmentId" IS NULL`
-            }`
+          ? ` INNER JOIN "organizationSegmentsAgg" osa ON osa."organizationId" = o.id AND osa."segmentId" IN ($(segments:csv))`
```
epipav, line 1622: I don't think this is going to work - We should still send a single segmentId to this query because we keep each subprojects, projects and project group aggregates in a single row in `organizationSegmentsAgg`

services/libs/data-access-layer/src/members/base.ts
```diff
@@ -229,7 +234,7 @@ export async function queryMembersAdvanced(
       FROM members m
       ${
         withAggregates
-          ? ` INNER JOIN "memberSegmentsAgg" msa ON msa."memberId" = m.id AND msa."segmentId" = $(segmentId)`
+          ? ` INNER JOIN "memberSegmentsAgg" msa ON msa."memberId" = m.id AND msa."segmentId" IN ($(segments:csv))`
```
epipav, line 237: same issue here

epipav (changes requested):
added a NIT and an issue

skwowet:
Closing this PR, since the underlying problem is different, and the fix is probably not related to it.

The code already supports projectGroup, project, and subproject segment IDs, but due to some issues with Argo segment mapping, it is not working as expected.

PR #2998, opened 2025-04-16
Prevent OpenSearch nested clause errors in organization merge suggestions

skwowet:
Prevents `too_many_nested_clauses` errors in OpenSearch when handling organizations with 150+ identities. It's similar to member suggestions but adjusted:

- Deduplicates organization identities by platform/value.
- Prioritizes verified identities by sorting.
- Limits processed identities to 140, reducing query complexity.

It fixes the error while keeping key search logic: verified identities match others' unverified (and vice-versa). Unverified identities are still considered but get deprioritized by sorting/slicing if the limit's hit.


skwowet:
@epipav I've updated the PR with the logic we went over. Thanks! ✌🏻

PR #3002, opened 2025-04-18
feat: jira integration

mbani01:
# Changes proposed ✍️

### What
Migrate Jira migration to Nango

### Why
The current Jira implementation is broken

### How
**BACKEND** (@skwowet & @epipav please check backend changes): 

Preparing API to use nango for Jira, using their 3 integrations:
1. jira-basic (Cloud)
2. jira-data-center-basic
3. jira-data-center-api-key

PS: I had to adjust some core functions to support the mapping between our Jira and Nango's Jira. This is because Nango treats each auth flow as a separate integration, while we have a single Jira integration that supports all those auth methods.

**FRONTEND** (@Sameh16 & @gaspergrom please check frontend changes):

1. Enable Jira to allow new connections (I just reverted the old implementation before disabling Jira)
2. Remove Jira from the warning message in the banner and keep Confluence only 

## Checklist ✅

### DIFF
--- a/backend/src/services/integrationService.ts
+++ b/backend/src/services/integrationService.ts
@@ -16,7 +16,8 @@ import {
 import {
   NangoIntegration,
   connectNangoIntegration,
-  createNangoIntegration,
+  createNangoConnection,
+  createNangoGithubConnection,
   deleteNangoConnection,
   getNangoConnectionData,
   getNangoConnections,
@@ -65,7 +66,6 @@ import { encryptData } from '../utils/crypto'
 
 import { IServiceOptions } from './IServiceOptions'
 import { getGithubInstallationToken } from './helpers/githubToken'
-import { jiraIntegrationData } from '../types/jiraTypes'
 
 const discordToken = DISCORD_CONFIG.token || DISCORD_CONFIG.token2
 
--- a/backend/src/services/integrationService.ts
+++ b/backend/src/services/integrationService.ts
@@ -23,7 +23,6 @@ import {
   getNangoConnections,
   setNangoMetadata,
   startNangoSync,
-  getNangoCloudSessionToken,
 } from '@crowd/nango'
 import { RedisCache } from '@crowd/redis'
 import { Edition, PlatformType } from '@crowd/types'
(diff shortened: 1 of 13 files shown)


services/libs/nango/src/types.ts
```diff
@@ -20,7 +22,23 @@ export const NANGO_INTEGRATION_CONFIG = {
...
+    },
+    syncs: {
+      ISSUES: 'issues',
+      ISSUE_COMMENT: 'issue-comments',
+      ISSUE_ATTACHMENTS: 'issue-attachments',
+    },
+  },
+  [NangoIntegration.JIRA_DATA_CENTER_BASIC]: {
+    models: {},
+    syncs: {},
```
themarolt, line 39: if there are no syncs here for these two integrations nothing will get started in `integrationService#1754` where startSync is called.

mbani01: Yes, I left them empty until we have all the integrations (jira-data-center-basic, jira-data-center-api-key) on [nango.yaml](https://github.com/CrowdDotDev/nango-integrations/blob/main/nango-integrations/nango.yaml) to avoid any potential confilicts

Sameh16 (approved):
LGTM for UI part

backend/src/services/integrationService.ts
```diff
@@ -1618,35 +1619,119 @@ export default class IntegrationService {
...
    * @param integrationData  to create the integration object
    * @returns integration object
+   * @remarks
+   * Supports the following authentication methods:
+   * 1. Jira Cloud (basic auth): Requires URL, username, and password (API key)
+   * 2. Jira Data Center (PAT): Requires URL and optionally a Personal Access Token
+   * 3. Jira Data Center (basic auth): Requires URL, username, and password (API key)
    */
-  async jiraConnectOrUpdate(integrationData) {
+  async jiraConnectOrUpdate(integrationData: jiraIntegrationData) {
```
epipav, line 1631: nit: `jiraIntegrationData` -> `JiraIntegrationData`, let's have types start with upper case always

epipav (commented):
added one nitpick, other than that lgtm :+1:

backend/src/services/integrationService.ts
mbani01: Good catch!

skwowet (approved):

services/libs/common/src/i18n/en.ts
```diff
@@ -152,6 +152,12 @@ const en = {
     git: {
       noIntegration: 'The Git integration is not configured.',
     },
+    jira: {
+      invalidUrl:
```
gaspergrom, line 156: We dont use translations anymore so in further prs you can also include directly in code.

gaspergrom (approved):

borfast (approved):

PR #3041, opened 2025-05-07
refactor: aggregate calculations for members and organizations

skwowet:
# Changes Proposed ✍️

This PR introduces a significant refactor and improvement to how we calculate and store member and organization aggregates, separating the calculation of core and display aggregates for better performance, maintainability, and future extensibility.

- **Refactor core aggregates calculation:**  
  - `syncMember` and `syncOrganization` now use the `activityRelations` table (instead of QuestDB) to calculate core aggregates such as `activeOn`, `activityCount`, and (for organizations) `memberCount`.  
  - This change improves query performance and reliability.

- **Async display aggregates calculation:**  
  - Introduces new async Temporal workflows to refresh/calculate display aggregates (`activeDaysCount`, `averageSentiment`, `activeOn`, `activityTypes`, etc.) for both members and organizations.
  - These jobs run daily, processing only recently indexed entities (tracked via the `indexed_entities` table and a persisted watermark in tenant settings).
  - Display aggregates are fetched from QuestDB (with plans to migrate to Tinybird in the future) and update the corresponding records in `memberSegmentsAgg` and `organizationSegmentsAgg`.
  - Default values for display aggregates are set during the initial sync to avoid database "not null" constraint issues, and are later updated by the async job.

## Other Improvements

- Added indexes on `activityRelations` to optimize core aggregate queries.
- Added `createdAt` and `updatedAt` to `organizationSegmentsAgg` to align with `memberSegmentsAgg`, and backfilled `updatedAt` in `memberSegmentsAgg` using `createdAt`.
- Removed redundant `sync-all-*` scripts from `search-sync-worker` (migrated to [`script-executor-worker`](https://github.com/CrowdDotDev/crowd.dev/blob/a190f72a283ff3bb5dc89d6b3f206243a13190bd/services/apps/script_executor_worker/src/workflows/sync)).
- Cleaned up redundant indexes on the `indexed_entities` table.
- Removed unused type definitions and functions for better code clarity.

## Overview of the Aggregation Now

- **Core Aggregates Calculation (sync):**  
  - When a member or organization profile changes (e.g., merge, unmerge, etc.), core aggregates are calculated and stored immediately, with placeholder defaults for display aggregates.
  - Entities are marked as indexed in `indexed_entities`.

- **Display Aggregates Calculation (async):**  
  - On a daily schedule, the async jobs fetch recently indexed entities (using a persisted `lastSyncedAt` watermark).
  - For each entity, display aggregates are calculated and updated in the DB.
  - Once all new entities are processed, the watermark is updated for the next run.

- **Why This Approach?**  
  - This separation allows us to keep core data fresh in real-time, while heavier analytical calculations are handled async, reducing load and improving reliability.
  - The new structure also prepares us for future migration of display aggregate calculations to Tinybird.


services/apps/profiles_worker/src/workflows/organization/refreshOrganizationDisplayAggregates.ts
```diff
@@ -0,0 +1,56 @@
...
+    return
+  }
+
+  const lastRow = result[result.length - 1]
+
+  for (const organization of result) {
+    const organizationDisplayAggregates = await getOrganizationDisplayAggregates(
+      organization.entity_id,
+    )
+    // todo:nathan test the changes with console log and testRun
```
themarolt, line 44: Another TODO here

services/apps/profiles_worker/src/workflows/organization/refreshOrganizationDisplayAggregates.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const lastUuid: string = args.lastUuid ?? null
+  let lastSyncedAt: string = args.lastSyncedAt ?? null
+
+  if (!lastSyncedAt) {
+    lastSyncedAt = await getLastOrganizationDisplayAggsSyncedAt()
+  }
+
+  // todo:nathan rm this after testing
+  console.log('lastSyncedAt', lastSyncedAt)
```
themarolt, line 29: Todo here

services/libs/data-access-layer/src/members/segments.ts
```diff
@@ -1,10 +1,16 @@
+import lodash from 'lodash'
```
themarolt, line 1: Don't import entire lodash just use individual npm packages like `lodash.pickby`.

services/libs/data-access-layer/src/organizations/segments.ts
```diff
@@ -1,10 +1,12 @@
+import lodash from 'lodash'
```
themarolt, line 1: Same here.

services/libs/opensearch/src/repo/indexing.repo.ts
```diff
@@ -69,4 +69,28 @@ export class IndexingRepository extends RepositoryBase<IndexingRepository> {
...
+  public async getRecentlyIndexedEntities(
+    entityType: IndexedEntityType,
+    batchSize: number,
+    lastSyncedAt: string,
+    lastUuid?: string,
+  ): Promise<IRecentlyIndexedEntity[]> {
+    return this.db().any(
+      `
+        SELECT "entity_id", "indexed_at"
+        FROM "indexed_entities"
```
themarolt, line 82: `index_entities` was mainly used to force reindex from scratch to opensearch and AFAIK it stopped being performant enough a while back - are you sure we should still be using this table? It's very bloated with ids from different big tables.

services/libs/opensearch/src/service/organization.sync.service.ts
```diff
@@ -346,6 +305,8 @@ export class OrganizationSyncService {
         }
       }
 
+      await this.indexingRepo.markEntitiesIndexed(IndexedEntityType.ORGANIZATION, organizationIds)
```
themarolt, line 308: This is probably where it will get slow over time with `indexed_entities`. Just wanted to make sure we have considered this since I never put much attention to the table with the indexes and queries... It worked before when we had ~20 mil. activities and 300k members :D

themarolt (changes requested):

themarolt:
Some nitpicks and questions but other than that code looks good - nice job!

services/apps/profiles_worker/src/main.ts
```diff
@@ -37,5 +41,9 @@ setImmediate(async () => {
 
   await scheduleRecalculateAffiliationsOfNewRoles()
 
+  // todo:nathan uncomment this after testing workflow
```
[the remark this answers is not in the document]
skwowet: Yeah, I’ll rm those `todo:nathan` blocks after testing in staging.

backend/src/database/migrations/V1746535447__activityRelationsIndexesForCoreAggs.sql
```diff
@@ -0,0 +1,36 @@
+-- Timestamps for incremental updates of member/org display aggregates
+alter table "tenants"
```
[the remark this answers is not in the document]
skwowet: Humm, good point! 🤔 

I’ll see if we can reuse any tables to store these values. If not, I’ll create a new one. Sound good?

services/libs/data-access-layer/src/members/segments.ts
skwowet: Will replace the `lodash` import with specific individual pkgs 👍🏻

services/libs/opensearch/src/repo/indexing.repo.ts
skwowet: I was initially planning to use the `memberSegmentsAgg` and `organizationSegmentsAgg` tables and basically use the `createdAt` column from those, running a similar select query to get recently synced items.

Something like this:
```sql
SELECT DISTINCT ON ("memberId") "memberId", "createdAt"
FROM "memberSegmentsAgg"
WHERE ("createdAt", "memberId") > ($(lastSyncedAt), $(lastMemberId))
ORDER BY "memberId", "createdAt" ASC
LIMIT $(batchSize);
```

I added a couple of indexes to the segment aggs table and checked the perf, and it looked good. 

But later, I noticed the `indexed_entities` table and saw we weren’t using it much other than for scripts to sync stuff from where it left off. I used this instead (added indexes to help the query, and it looks good), so we can offload some operations from the aggs table.

Wdyt? Should we stick with the segment aggs table? 🤔

backend/src/database/migrations/V1746535447__activityRelationsIndexesForCoreAggs.sql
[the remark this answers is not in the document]
themarolt: Yep perfect.

services/libs/opensearch/src/repo/indexing.repo.ts
themarolt: We can use `indexed_entities` but perhaps make sure that it has correct indexes :D

services/libs/opensearch/src/service/organization.sync.service.ts
skwowet: Yep, makes sense.

We don’t have many indexes or queries on this table, so I think we're good for now. But yeah, we should definitely look into this more before it starts slowing down over time.

services/apps/profiles_worker/src/activities.ts
```diff
@@ -17,4 +31,16 @@ export {
   syncMember,
   syncOrganization,
   findMembersInOrganization,
+  // -
```
epipav, line 34: nit: empty comment here

services/apps/profiles_worker/src/activities.ts
```diff
@@ -17,4 +31,16 @@ export {
...
+  getMemberDisplayAggsLastSyncedAt,
+  touchMemberDisplayAggsLastSyncedAt,
+  getMembersForDisplayAggsRefresh,
+  getMemberDisplayAggregates,
+  setMemberDisplayAggregates,
+  // -
```
epipav, line 40: same

epipav (commented):
Looks great 👍 I left a small comment regarding the commented-out code.

Could u add somewhere the difference between core and display aggregates? Perhaps this could be added in the new workflows section of the profiles worker—might be a helpful reference point.

themarolt (approved):

PR #3046, opened 2025-05-09
feat: refactor display of collections and insights projects [INS-469]

borfast:
[Ticket in Linear](https://linear.app/lfx/issue/INS-469/refactor-display-of-collections-and-insights-projects).

This adds a 'starred' column to the 'collections' table, as well as the corresponding TinyBird schema changes.


epipav (commented):
looks good :+1:  

There's already a starred field under `collectionsInsightsProjects` table. If this was the intended place for it (collections), maybe we can drop that one since it's not used anywhere

Sameh16:
> looks good 👍
> 
> There's already a starred field under `collectionsInsightsProjects` table. If this was the intended place for it (collections), maybe we can drop that one since it's not used anywhere

@epipav  we still need this one too

borfast:
> looks good 👍
> 
> There's already a starred field under `collectionsInsightsProjects` table. If this was the intended place for it (collections), maybe we can drop that one since it's not used anywhere

I believe they're both in use.

PR #3057, opened 2025-05-12
feat: on TinyBird, when sorting collections by "starred", sort them also by "projectCount"

borfast:
On TinyBird, when sorting collections by "starred", we also want to sort them by "projectCount".

This is just a quick fix for now. Ideally we would allow sorting on multiple fields but we will tackle that later if and when the need arises.


borfast:
This was created on a wrong branch. I'm closing it and reopening with the right branch.

PR #3111, opened 2025-05-26
refactor: use `activityRelations` for affiliations, merges, and others

skwowet:


joanagmaia:
@skwowet I believe there is one place that we will also have to update:
- When we process an incoming activity, we check the existing memberId from the activities table. Since we will stop updating these in activities, we should update that method to start reading the memberId from activityRelations.
Probably better to sync with @themarolt on this one, and make sure that there isn't anywhere else that we need to update.

skwowet:
@joanagmaia Thanks for the heads-up! 👍

I’ll refactor it to read from `activityRelations` and double-check with @themarolt as well.

services/apps/profiles_worker/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@crowd/profiles-worker",
   "scripts": {
-    "start": "CROWD_TEMPORAL_TASKQUEUE=profiles SERVICE=profiles-worker tsx src/main.ts",
+    "start": "CROWD_TEMPORAL_TASKQUEUE=profiles SERVICE=profiles-worker tsx src/main.ts LOG_LEVEL=trace",
```
themarolt: This is used on kubernetes on production so I wouldn't keep this `LOG_LEVEL=trace` there to avoid spamming logs.

services/libs/data-access-layer/src/activityRelations/index.ts
```diff
@@ -1,7 +1,170 @@
...
+  const columnString = columns.map((c) => `ar."${c}"`).join(', ')
+
+  let query = `
+    select ${columnString}
+    ${baseQuery}
+    order by ${orderByString}
+  `
+
+  if (!arg.noLimit || arg.limit > 0) {
+    query += ` limit $(limit) offset $(offset)`
```
themarolt, line 148: AFAIK limit and offset were different on questdb - if this function was copy pasted for postgres activityRelations instead of questdb activities it may have some hidden issues with SQL since questdb and postgresql SQL is not the same in a lot of cases...

services/apps/profiles_worker/package.json
skwowet: Yeah yeah, I added it to check if everything looks good after deployment. Didn't update the env to avoid spamming logs to other services.

services/libs/data-access-layer/src/activityRelations/index.ts
skwowet: Yep, I refactored the func manually instead of copy-pasting, due to compatibility issues like you mentioned.

We can actually simplify/improve the func further, but I kept it as a drop-in replacement for `queryActivities` (for places where we don't need the actual activity data), so didn’t make any improvements to avoid breaking changes.

services/apps/profiles_worker/package.json
skwowet: Removed `LOG_LEVEL=trace` ✅

PR #3118, opened 2025-05-27
fix: prevent entity-merging-worker pod from crashing

skwowet:


services/libs/data-access-layer/src/activities/update.ts
```diff
@@ -19,49 +19,44 @@ export async function streamActivities(
...
-    [],
-    {
-      batchSize: 1000,
-      highWaterMark: 250,
-    },
-  )
+  const sqlQuery = `SELECT * FROM activities WHERE "deletedAt" is null and ${whereClause}`
 
+  // eslint-disable-next-line
+  const qi = createQueryIterable(qdb as any)
```
skwowet, line 25: @themarolt how can we pass the client/pool here? 🤔

skwowet:
Prepared a fix in a different PR: #3124 

So closing this one!

PR #3135, opened 2025-05-30
feat: add the cocomo analysis tool [INS-435]

borfast:
This adds the COCOMO analysis tool for getting the software value for projects.

I'm still rounding off a couple of corners, so I'm not opening this as a full PR yet, but feel free to provide feedback if you have any.


themarolt:
The code looks fine but I don't understand why this is in crowd.dev repo? It doesn't have anything to do with CM right? and if it's a service that will live in CM kubernetes cluster perhaps just move it to `services/apps/software_value` folder just so it's not in root confusing people if it's just one more service.

borfast:
> The code looks fine but I don't understand why this is in crowd.dev repo? It doesn't have anything to do with CM right?

@joanagmaia said I should put it in this repository because the crowd-kube one is more for configuration and not code. Initially I thought we would put it in a separate repository, since it's its own thing. The only "point of contact" with CM is the fact that this has to access the CM database. Perhaps we can rethink this. What do you think?

> and if it's a service that will live in CM kubernetes cluster perhaps just move it to `services/apps/software_value` folder just so it's not in root confusing people if it's just one more service.

It's not going to run in the K8s cluster. Maybe I should have been more explicit about that in the PR description, sorry. It's meant to be a standalone service running in the git-integration machine, and it doesn't reuse anything of the current code base.

In retrospect, putting it in the root of the repository is not the best idea indeed; I think this was just an idea that was mentioned at the beginning of the project and ended up stuck in my head. But you're absolutely right, the root of the repo is not the right place for it. I'll move it.

themarolt:
I would suggest just moving it to the git integration repo and it should be fine. IMHO it's confusing to have it in CM repo.

joanagmaia:
@borfast @themarolt 
My suggestion on adding it to the crowd.dev repo was because I thought this was adding the data into CM's postgres.
But I believe it's even a new DB specifically for this, also didn't know it was written in go. So perhaps yeah it does not make sense to have it in crowd.dev.

Since this is directly interacting with the git-integration, then yeah we an also add it there. That way it's easier to know what it affects, and Mouad can also check it out easily when he plans for the git integration refactor.

borfast:
@joanagmaia, my apologies, I thought you were aware of the details (a separate DB and written in Go).

I'll close this PR, then, and push the code to the git-integration repository.

PR #3173, opened 2025-06-12
fix: use segmentId instead of tenantId to get linkedIn tokens from nango [CM-335]

mbani01:
# Changes proposed ✍️
Use `segmentId` instead of `tenantId` to get LinkedIn tokens from nango

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/backend/src/services/integrationService.ts
+++ b/backend/src/services/integrationService.ts
@@ -5,7 +5,7 @@ import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
 import lodash from 'lodash'
 import moment from 'moment'
 
-import { DEFAULT_TENANT_ID, EDITION, Error400, Error404, Error542 } from '@crowd/common'
+import { EDITION, Error400, Error404, Error542 } from '@crowd/common'
 import {
   NangoIntegration,
   connectNangoIntegration,
@@ -846,8 +846,8 @@ export default class IntegrationService {
     throw new Error404(this.options.language, 'errors.linkedin.cantOnboardWrongStatus')
   }
 
-  async linkedinConnect() {
-    const nangoId = `${DEFAULT_TENANT_ID}-${PlatformType.LINKEDIN}`
+  async linkedinConnect(segmentId: string) {
+    const nangoId = `${segmentId}-${PlatformType.LINKEDIN}`
 
     let token: string
     try {
--- a/backend/src/api/integration/helpers/linkedinConnect.ts
+++ b/backend/src/api/integration/helpers/linkedinConnect.ts
@@ -4,6 +4,7 @@ import PermissionChecker from '../../../services/user/permissionChecker'
 
 export default async (req, res) => {
   new PermissionChecker(req).validateHas(Permissions.values.tenantEdit)
-  const payload = await new IntegrationService(req).linkedinConnect()
+  const segmentId = req.body.segments[0]
+  const payload = await new IntegrationService(req).linkedinConnect(segmentId)
   await req.responseHandler.success(req, res, payload)
 }


services/apps/cron_service/src/jobs/queueMonitoring.job.ts
mbani01, line 1: Although the PR shows many changes, the only actual modification is on [line 6 (unused variables)](https://github.com/CrowdDotDev/crowd.dev/blob/099c064ea43454b0f21eab89adc3467b201c5e59/services/apps/cron_service/src/jobs/queueMonitoring.job.ts#L6C10-L6C22), which slipped through without lint validation. All other changes are just noise from outdated commit history — no functional edits were made

backend/src/services/integrationService.ts
```diff
@@ -846,8 +846,8 @@ export default class IntegrationService {
     throw new Error404(this.options.language, 'errors.linkedin.cantOnboardWrongStatus')
   }
 
-  async linkedinConnect() {
-    const nangoId = `${DEFAULT_TENANT_ID}-${PlatformType.LINKEDIN}`
+  async linkedinConnect(segmentId: string) {
+    const nangoId = `${segmentId}-${PlatformType.LINKEDIN}`
```
themarolt, line 850: In this file:
https://github.com/CrowdDotDev/crowd.dev/blob/099c064ea43454b0f21eab89adc3467b201c5e59/services/apps/integration_stream_worker/src/service/integrationStreamService.ts#L456

we are still using `${DEFAULT_TENANT_ID}-${streamInfo.integrationType}` when we pass this down the code to the nango function like this: https://github.com/CrowdDotDev/crowd.dev/blob/099c064ea43454b0f21eab89adc3467b201c5e59/services/libs/integrations/src/integrations/linkedin/api/organization.ts#L12

So I'm not sure how this will work later on when the integration is actually being processed as there won't be the same `connectionId` passed when creating the nango connection vs when we try to get the auth token from nango.

mbani01: Now `nangoId` is stored in `integration.settings`, to be used later for stream processing

services/apps/cron_service/src/jobs/queueMonitoring.job.ts
```diff
@@ -3,7 +3,6 @@ import CronTime from 'cron-time-generator'
 import { IS_PROD_ENV, distinct } from '@crowd/common'
 import { Logger } from '@crowd/logging'
 import { KafkaAdmin, QUEUE_CONFIG, getKafkaClient } from '@crowd/queue'
```
themarolt: I think you can just merge main to fix this file.

services/apps/integration_run_worker/src/service/integrationRunService.ts
```diff
@@ -295,13 +295,16 @@ export default class IntegrationRunService extends LoggerBase {
     const cache = new RedisCache(`int-${runInfo.integrationType}`, this.redisClient, this.log)
 
     const nangoConfig = NANGO_CONFIG()
-
+    const nangoId =
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (runInfo.integrationSettings as any)?.nangoId ||
+      `${DEFAULT_TENANT_ID}-${runInfo.integrationType}`
```
themarolt: Isn't it broken to use `tenantId` now?

So to explain:
- when crowd.dev repo was still under crowd.dev company we had multiple customers (tenants) and each could connect only one integration per platform so for example only 1 linkedin integration could be connected
- but now we don't have multiple tenants (there are still tenants but it's basically deprecated feature - only have one row in tenants db table) but we can have multiple linkedins for this same default tenant id - but only one per segment (project)

so I would hard deprecate this nango id with default tenant id as it doesn't make sense at all even for legacy connections.

themarolt: Are existing nango connection ids gonna be updated manually or with a migration?

mbani01: I agree, I kept the default value just to avoid hiccups before updating db, because now we have a mix between segmentId (for new LinkedIn connections) and tenantId for other integrations.
But yes in any ways we shouldn't have any tenantId after the fix. I'll remove it.
Since we have less than 10 connections, I'll manually update them as it's a one time fix

PR #3178, opened 2025-06-13
feat: add the TInyBird resources for the search volume data

borfast:
[Ticket in Linear](https://linear.app/lfx/issue/INS-658/implement-the-data-pipeline-to-get-the-data-into-tinybird)

This adds the TinyBird data source and pipe with the respective endpoint for the search volume data, to be consumed by the Insights application.

### DIFF
--- a/services/libs/tinybird/datasources/searchVolume.datasource
+++ b/services/libs/tinybird/datasources/searchVolume.datasource
@@ -1,0 +1,12 @@
+SCHEMA >
+    `id` UInt64 `json:$.record.id`,
+    `projectId` UUID `json:$.record.project_id`,
+    `keyword` String `json:$.record.keyword`,
+    `dataTimestamp` DateTime64(3) `json:$.record.data_timestamp`,
+    `volume` UInt64 `json:$.record.volume`,
+    `updatedAt` DateTime64(3) `json:$.record.updated_at`
+
+ENGINE ReplacingMergeTree
+ENGINE_PARTITION_KEY toYear(updatedAt)
+ENGINE_SORTING_KEY projectId, keyword
+ENGINE_VER updatedAt
--- a/services/libs/tinybird/datasources/searchVolume.datasource
+++ b/services/libs/tinybird/datasources/searchVolume.datasource
@@ -1,12 +1,12 @@
 SCHEMA >
     `id` UInt64 `json:$.record.id`,
-    `projectId` UUID `json:$.record.project_id`,
-    `keyword` String `json:$.record.keyword`,
+    `insightsProjectId` UUID `json:$.record.insights_project_id`,
+    `slug` String `json:$.record.slug`,
     `dataTimestamp` DateTime64(3) `json:$.record.data_timestamp`,
     `volume` UInt64 `json:$.record.volume`,
     `updatedAt` DateTime64(3) `json:$.record.updated_at`
 
 ENGINE ReplacingMergeTree
 ENGINE_PARTITION_KEY toYear(dataTimestamp)
-ENGINE_SORTING_KEY projectId
+ENGINE_SORTING_KEY dataTimestamp
 ENGINE_VER updatedAt
--- a/services/libs/tinybird/datasources/searchVolume.datasource
+++ b/services/libs/tinybird/datasources/searchVolume.datasource
@@ -8,5 +8,5 @@ SCHEMA >
 
 ENGINE ReplacingMergeTree
 ENGINE_PARTITION_KEY toYear(dataTimestamp)
-ENGINE_SORTING_KEY dataTimestamp
+ENGINE_SORTING_KEY slug, dataTimestamp
 ENGINE_VER updatedAt
--- a/services/libs/tinybird/datasources/searchVolume.datasource
+++ b/services/libs/tinybird/datasources/searchVolume.datasource
@@ -7,6 +7,6 @@ SCHEMA >
     `updatedAt` DateTime64(3) `json:$.record.updated_at`
 
 ENGINE ReplacingMergeTree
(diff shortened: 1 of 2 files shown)


services/libs/tinybird/datasources/searchVolume.datasource
```diff
@@ -0,0 +1,12 @@
+SCHEMA >
+    `id` UInt64 `json:$.record.id`,
+    `projectId` UUID `json:$.record.project_id`,
```
epipav: NIT: can we call this `insightsProjectId` to be consistent with the rest?

services/libs/tinybird/pipes/search_volume.pipe
```diff
@@ -0,0 +1,32 @@
+DESCRIPTION >
+    Serves the project search volume data for the `Search queries` widget
+
+NODE searchVolume_pipe
+SQL >
+    %
+    SELECT projectId, keyword, dataTimestamp, volume, updatedAt
```
epipav: NIT: same naming nitpick as above

services/libs/tinybird/datasources/searchVolume.datasource
```diff
@@ -0,0 +1,12 @@
...
+    `keyword` String `json:$.record.keyword`,
+    `dataTimestamp` DateTime64(3) `json:$.record.data_timestamp`,
+    `volume` UInt64 `json:$.record.volume`,
+    `updatedAt` DateTime64(3) `json:$.record.updated_at`
+
+ENGINE ReplacingMergeTree
+ENGINE_PARTITION_KEY toYear(dataTimestamp)
+ENGINE_SORTING_KEY projectId
```
epipav: We can add `dataTimestamp` to the sorting key - since queries will filter out first by insightsProjectId and timestamp, I'm guessing it's gonna help with filtering performance

services/libs/tinybird/pipes/search_volume.pipe
```diff
@@ -0,0 +1,32 @@
...
+    FROM searchVolume
```
epipav: You probably want `FINAL` here since the engine is `ReplacingMergeTree` in the source

services/libs/tinybird/pipes/search_volume.pipe
```diff
@@ -0,0 +1,32 @@
...
+            AND searchVolume.dataTimestamp
+            <= {{
+                DateTime(
+                    endDate,
+                    description="Select records with dataTimestamp only before the given date",
+                    required=False,
+                )
+            }}
+        {% end %}
+        {% if defined(projectId) %} AND searchVolume.projectId = {{ projectId }} {% end %}
```
epipav: NIT: It's gonna help with consistency and usability of the pipe if we provide a project slug filter instead of an opaque id filter here. (since everywhere else FE expects this filtering to be done by slug) 

There's a reusable example [here](https://github.com/CrowdDotDev/crowd.dev/blob/e8dc098313ccef3e6a0d6f71fe0b84345e640bd4/services/libs/tinybird/pipes/package_downloads.pipe)
Another benefit of using `segments_filtered` pipe for this is that it provides `insightsProjects.enabled` check as well

epipav (changes requested):
added a few nitpicks and some consistency points

services/libs/tinybird/datasources/searchVolume.datasource
borfast: Sure, no problem.

services/libs/tinybird/datasources/searchVolume.datasource
borfast: Good point. In fact, the project ID will probably not help much here at all, since we're going to query by keyword (the project slug), not project ID, so I think removing the project ID from here is even better.

services/libs/tinybird/pipes/search_volume.pipe
borfast: The `keyword` filter below this one is actually the slug. I'll change the name, it doesn't make sense to use the syntax from Keywords Everywhere here.

borfast:
@epipav , can you review this again, please?

services/libs/tinybird/pipes/search_volume.pipe
```diff
@@ -0,0 +1,32 @@
...
+            AND searchVolume.dataTimestamp
+            <= {{
+                DateTime(
+                    endDate,
+                    description="Select records with dataTimestamp only before the given date",
+                    required=False,
+                )
+            }}
+        {% end %}
+        {% if defined(insightsProjectId) %} AND searchVolume.insightsProjectId = {{ insightsProjectId }} {% end %}
```
epipav: If you apply the filter like this here
`insightsProjectId = (select insightsProjectId from segments_filtered)`, it'll also check if the project is enabled. r. Currently, it doesn't check that, and if the project is disabled, data will be returned even though the project is disabled.

epipav (approved):
looks good, added one improvement to the insightsProjectId filter for convenience

services/libs/tinybird/pipes/search_volume.pipe
borfast: Good one, I'll add it. Thanks! :+1:

borfast: I ended up forgetting to add this. Just to double check, you meant adding this, right?
`AND searchVolume.insightsProjectId IN (select insightsProjectId from segments_filtered)`

The difference being using `IN` instead of `=`.

Or did you mean something else?

borfast:
As discussed in Slack, I made the change to the filter and I also renamed slug to project, to match the rest of the pipes and data sources.

epipav (approved):

PR #3193, opened 2025-06-17
fix: implement pagination for getOrgRpos using octokit [CM-2361]

mbani01:
# Changes proposed ✍️
Add  pagination support for getOrgRpos using octokit

### What
copilot:summary
​
copilot:poem

### Why
Currently, we're getting only 30 repos (max) for each organization as it's lacking pagination for that endpoint

### How
copilot:walkthrough

## Checklist ✅


gaspergrom (approved):

backend/src/services/githubIntegrationService.ts
```diff
@@ -57,14 +58,17 @@ export default class GithubIntegrationService {
...
+    const octokit = new Octokit({
+      auth: `Bearer ${token}`,
+    })
+
+    const repos = await octokit.paginate(octokit.rest.repos.listForOrg, {
       org,
-      headers: {
-        authorization: `bearer ${auth}`,
-      },
+      per_page: 100, // max results per page is 100
```
themarolt, line 68: We might as well do a proper pagination right? As in check if there are more pages and load all in a loop until no more pages?

mbani01: `octokit.paginate` is actually doing it, so even if we set `per_page` to 1, we'll still get all the pages.
I set it to the max to reduce the number of iterations

themarolt: Ah ok then no worries!

themarolt (approved):

PR #3196, opened 2025-06-18
fix(member): preserve manual changes in `setAttributesDefaultValues`

skwowet:


skwowet:
Closing this PR since the issue’s been fixed the right way:

- `isBot.custom` key wasn’t being set, so default kept getting overridden by incoming data. 
- Fixed the data by adding `isBot.custom: true` where needed.
- Frontend now correctly sends custom instead of just default (see [#3198](https://github.com/your-repo/your-project/pull/3198))
- Since `custom` has highest priority, the original logic works fine - no workaround needed.

PR #3199, opened 2025-06-19
feat: add segments GitHub insights api

ulemons:
# Changes proposed ✍️

### What:
Adding api for fetching github insights

### Why:
We want to add an api in order to preview github information when adding a new Insights project from a segment

### How
1. adding api entry point: `/segments/:id/github-insights`
2. implementing the service: `findGithubInsightsForSegment`
3. We use LLM to understand which is the main oganization if a segment has more than one
4. We use LLM to retrieve a generic description of the project
5. Other static info are taken from github

#### additional-info:
doc for the interagion with github can be found [here](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28).

📌 Linear task: https://linear.app/lfx/issue/CM-2314/get-static-information-for-projects-from-github


backend/src/services/collectionService.ts
```diff
@@ -464,4 +465,46 @@ export class CollectionService extends LoggerBase {
...
+    const githubIntegrations = integrations.filter(
+      (integration) =>
+        integration.platform === PlatformType.GITHUB ||
+        integration.platform === PlatformType.GITHUB_NANGO,
+    )
+
+    const result = []
+
+    // FIXME: if we have just one github integration, we can avoid the loop
+    for (const { token, settings } of githubIntegrations) {
```
joanagmaia: We won't be able use this token all the time. This `token` is only available for integrations with platform 'github' and not for integrations platform 'github-nango'.

The reason is that for platform github we ask the user for specific permissions, which grants us access to a token with read permissions. Whereas in github-nango, we use a mechanism of token rotation where we don't require users permissions to access the orgs.

So instead of leveraging the token from integrations which might be null, we should use the github token from an app we already have available. We are already leveraging this in other places -> https://github.com/CrowdDotDev/crowd.dev/blob/main/backend/src/services/helpers/githubToken.ts#L9

backend/src/services/collectionService.ts
```diff
@@ -464,4 +465,46 @@ export class CollectionService extends LoggerBase {
...
+          twitter,
+          website,
+          topics,
+        })
+      }
+    }
+
+    // FIXME: investigate how to chose between multiple orgs result
+    return result[0] || {}
+  }
```
joanagmaia: What I believe we could do here where there are multiple orgs:
1. For all orgs, try to get the organization name and description from GitHub. So we could extend the method you already have in place to also get name and description.
2. Once we have all of this information for all organization I would create a prompt to run against Amazon bedrock (we already have a method in place to use, we would just need to prepare the prompt), to ask the LLM which organization is the main one from the project.

Why I think this could work? We currently have 3 integrations where we have more than 1 organization. For all of them, it was pretty clear based on the name of the organization in github which one was the main one. So we could test against the LLM to see if it returned what we expected for these 3 use cases.

backend/src/services/helpers/githubOrganization.ts
```diff
@@ -0,0 +1,74 @@
+import axios from 'axios'
+
+type OrganizationInfo = {
+  logoUrl: string
+  github: string
+  website: string | null
+  twitter: string | null
+}
+
+export const getGithubOrganization = async ({
```
joanagmaia: We already have this method available in another file https://github.com/CrowdDotDev/crowd.dev/blob/main/backend/src/services/githubIntegrationService.ts#L45
It's not returning everything we need but I believe we could simply adjust it to return extra fields. Also, perhaps it makes sense that these methods live in `githubIntegrationService` just because we are already using it to get data from the GitHub api. Let me know what you think!

backend/src/services/collectionService.ts
joanagmaia: We already have a LLMService in place https://github.com/CrowdDotDev/crowd.dev/blob/main/services/libs/common_services/src/services/llm.service.ts#L28

backend/src/services/helpers/githubOrganization.ts
ulemons: the api `/search/users` used in the already existing `findOrgs` is not returning the twitter account and the website, in order to have these info we should use instead the /orgs for each organization to retreive those detail. I could use the findOrgs for getting the name of the organizations, and then implement a new one like findOrgDetails for getting the sepcific information for each organization. WDYT ?

ulemons: 1. Put everything under `githubIntegrationService` is great for me.
2. Unfortunately, the current `/search/users` API does not return the website or Twitter account, which can only be retrieved using the specific `/orgs` API. 
3. It is not possible to retrieve the LinkedIn account from github.

I could add a `findOrgDetail` and loop over this, or use the `findOrgs` and we ll ignore the website and twitter.

I would prefer not to add the information in the current `findOrgs` because this would mean that we must loop over each organization in this function too, but probably we don't need this here. (let me know if my point was not clear) 
 
WDYT?

joanagmaia: @ulemons yeah you're absolutely right, it's a different api, didn't even notice. Yeah I agree that we can use the orgs api, and we can simply create a new method to handle the new request we want.

services/libs/types/src/llm.ts
```diff
@@ -82,6 +82,14 @@ export const LLM_SETTINGS: Record<LlmQueryType, ILlmSettings> = {
       temperature: 0,
     },
   },
+  [LlmQueryType.MATCH_MAIN_GITHUB_ORGANIZATION_AND_DESCRIPTION]: {
+    modelId: LlmModelType.CLAUDE_3_5_SONNET,
```
epipav: AFAIK V2 is only better and with the same pricing - why not use V2 here as well?

services/libs/common_services/src/services/llm.service.ts
```diff
@@ -17,26 +16,26 @@ import {
   LlmMemberEnrichmentResult,
   LlmQueryType,
 } from '@crowd/types'
-
-import { LlmPromptHistoryRepository } from '../repos/llmPromptHistory.repo'
+import { insertPromptHistoryEntry } from '../../../data-access-layer/src/prompt-history'
+import { QueryExecutor } from '../../../data-access-layer/src/queryExecutor'
```
epipav: It'll be cleaner and more maintainable if we import these with package imports
`import { insertPromptHistoryEntry, QueryExecutor } from '@crowd/data-access-layer'`
by installing the library in common_services

backend/src/services/githubIntegrationService.ts
```diff
@@ -73,4 +79,105 @@ export default class GithubIntegrationService {
...
+
+      analyze the project’s content (e.g., README, code, metadata) to:
+
+      1. Identify which organization from the array is the main one associated with the project. Return null if none is a clear match.
+      2. Generate a neutral, objective description of what the project does. Return null if no meaningful description can be inferred.
+
+      If no match is found:
+      Return: null
+
+      if a matche is found:
```
epipav: nit: typo here

backend/src/services/githubIntegrationService.ts
```diff
@@ -73,4 +79,105 @@ export default class GithubIntegrationService {
...
+    })
+
+    return {
+      description: data.description || null,
+      github: data.html_url,
+      logoUrl: data.avatar_url,
+      name: data.login,
+      twitter: data.twitter_username || null,
+      website: data.blog || null,
+    }
```
epipav: NIT: Could data be null/empty here?

backend/src/services/githubIntegrationService.ts
```diff
@@ -73,4 +79,105 @@ export default class GithubIntegrationService {
...
+      try {
+        const res = await request(`GET /repos/${org}/${repo.name}/topics`, {
+          headers: {
+            authorization: `bearer ${auth}`,
+          },
+        })
+
+        res.data.names.forEach((topic: string) => topicSet.add(topic))
+      } catch (err) {
+        console.error(`Failed to fetch topics for ${repo.name}:`, err.response?.data || err.message)
```
epipav: NIT: It would be better to use the logger coming from LoggerBase here instead of the console

backend/src/services/collectionService.ts
```diff
@@ -464,4 +466,48 @@ export class CollectionService extends LoggerBase {
 
     return result
   }
+
+  async findGithubInsightsForSegment(segmentId: string) {
```
epipav: NIT: Missing function return type here

epipav (commented):
Looks great! 👌 Just added a few small nitpicks. Also, would be nice to add explicit return types to functions where they're missing

backend/src/services/githubIntegrationService.ts
ulemons: if the organization is not found or something the api should throw an error, so data should never be null. By the way i could add a try catch for robustness

joanagmaia:
@ulemons can you just edit the name of the PR to match the commitlint convention

services/libs/types/src/llm.ts
```diff
@@ -82,6 +82,14 @@ export const LLM_SETTINGS: Record<LlmQueryType, ILlmSettings> = {
       temperature: 0,
     },
   },
+  [LlmQueryType.MATCH_MAIN_GITHUB_ORGANIZATION_AND_DESCRIPTION]: {
+    modelId: LlmModelType.CLAUDE_3_5_SONNET_V2,
```
skwowet, line 86: @joanagmaia We were planning to upgrade our LLM model usage from Sonnet 3.5 to Sonnet 4 for merge-related tasks. Should we also use Sonnet 4 in this case?

joanagmaia: Once we tackle that task we can probably test for the different use cases. Since Umberto already tested with 3.5 and it was ok, we could move forward with this for the time being

PR #3219, opened 2025-06-27
feat: add insight projects widgets api

ulemons:
# Changes proposed ✍️

### What:
We want to move the widgets logic in the backend. 
With a given project we want to understand what are the widgets related to its connected platforms.
We want to expose an api which gives back all the widget for a specific project.

### Why:
In this way when we will create an insight project automatically we can trigger the api and know what widget to add.

### How:
1. added a new api `/segments/:id/widgets` which retrieves the widget related to a segment.
2. Default widget values logic moved to `lib/types/src/enums/widget.ts`
3. The logic now relies on the existing type `PlatformType` from the common lib
4. A new typeguard was introduced in `services/libs/types/src/enums/platforms.ts` in order to provide better type checking.

## Checklist ✅


backend/src/api/collections/segmentsWidgetsGet.ts
```diff
@@ -0,0 +1,23 @@
+import { CollectionService } from '@/services/collectionService'
+
+import Permissions from '../../security/permissions'
+import PermissionChecker from '../../services/user/permissionChecker'
+
+/**
+ * GET /collections/insights-projects/:id/widgets
```
joanagmaia: Nitpick: Is the path on the documentation correct?

services/libs/types/src/enums/platforms.ts
```diff
@@ -82,3 +82,7 @@ export const integrationProfileUrl: Record<IntegrationType, (username: string) =
   [IntegrationType.DISCOURSE]: () => null,
   [IntegrationType.GIT]: () => null,
 }
+
+export function isValidPlatform(value: unknown): value is PlatformType {
+  return typeof value === 'string' && Object.values(PlatformType).includes(value as PlatformType)
+}
```
joanagmaia: Kind of a nitpick 😄 but these files are being used for enums only, so perhaps not the best place to add this function.

I would argue that if we're only using this in one place `collectionService`, we can simply add the function on that file as well.

backend/src/services/collectionService.ts
```diff
@@ -517,4 +517,31 @@ export class CollectionService extends LoggerBase {
...
+    const integrations = await fetchIntegrationsForSegment(qx, segmentId)
+    const platforms = [
+      ...new Set(integrations.map((integration) => integration.platform).filter(isValidPlatform)),
+    ]
+
+    for (const platform of platforms) {
+      Object.entries(DEFAULT_WIDGET_VALUES).forEach(([key, config]) => {
+        if (
+          config.enabled &&
+          config.platform.map((p) => p.toLowerCase()).includes(platform.toLowerCase())
```
joanagmaia: Optimization will be negligent, so definitely no need to change, just a suggestion:
- .some() returns true or false immediately on finding a match.
```
const platformIncludesWidget = config.platform.some(p => p.toLowerCase() === platform.toLowerCase());
```

services/libs/types/src/enums/platforms.ts
ulemons: you re right maybe if we will use this from other services in future we will think how to centralize

PR #3237, opened 2025-07-04
fix: don't initialise QuestDB automatically

borfast:
Due to a top-level variable being initialised in the data access layer, it was being executed immediately when the code was imported, which was causing issues when QuestDB was not in use.

This moves the variable initialistion to inside the function that uses it, so we don't have the above problem.

### DIFF
--- a/services/libs/data-access-layer/src/conversations/ilp.ts
+++ b/services/libs/data-access-layer/src/conversations/ilp.ts
@@ -17,7 +17,9 @@ export async function insertConversations(
   const now = Date.now()
 
   if (conversations.length > 0) {
-    ilp = getClientILP()
+    if (!ilp) {
+      ilp = getClientILP()
+    }
 
     for (const conversation of conversations) {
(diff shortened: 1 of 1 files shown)


services/libs/data-access-layer/src/conversations/ilp.ts
```diff
@@ -17,6 +17,8 @@ export async function insertConversations(
   const now = Date.now()
 
   if (conversations.length > 0) {
+    ilp = getClientILP()
```
themarolt: Everything ok except I would cache the already created instance and not create instance every time the function is called.

themarolt (changes requested):

services/libs/data-access-layer/src/conversations/ilp.ts
borfast: Good point :+1:  I changed it to check if the variable has been initialised first and only do so if it hasn't yet.

PR #3238, opened 2025-07-04
feat: reflect integrations on insights

ulemons:
# Changes proposed ✍️

### What
Whenever an integration is created, updated, or disconnected, we want to reflect those changes on the corresponding insightsProject, if one exists.

### How
1. A new function `findInsightsProjectsBySegmentId` has been added to retrieve the relevant `insightsProject` associated with a given segment ID.
2. A static method `isCodePlatform` has been introduced to determine whether a platform is a code platform.
3. The integration create/update callbacks now include logic to update the associated insightsProject.
4. On the destroyAll callback remove repositories and plugins from the insightsProject.

## Checklist ✅

### DIFF
--- a/backend/src/services/collectionService.ts
+++ b/backend/src/services/collectionService.ts
@@ -399,92 +399,94 @@ export class CollectionService extends LoggerBase {
   }
 
   async findRepositoriesForSegment(segmentId: string) {
-    const qx = SequelizeRepository.getQueryExecutor(this.options)
-    const integrations = await fetchIntegrationsForSegment(qx, segmentId)
+    return SequelizeRepository.withTx(this.options, async (tx) => {
+      const qx = SequelizeRepository.getQueryExecutor(this.options, tx)
+      const integrations = await fetchIntegrationsForSegment(qx, segmentId)
 
-    // Initialize result with platform arrays
-    const result: Record<string, Array<{ url: string; label: string }>> = {
-      git: [],
-      github: [],
-      gitlab: [],
-      gerrit: [],
-    }
-
-    const addToResult = (platform: PlatformType, fullUrl: string, label: string) => {
-      const platformKey = platform.toLowerCase()
-      if (!result[platformKey].some((item) => item.url === fullUrl)) {
-        result[platformKey].push({ url: fullUrl, label })
+      // Initialize result with platform arrays
+      const result: Record<string, Array<{ url: string; label: string }>> = {
+        git: [],
+        github: [],
+        gitlab: [],
+        gerrit: [],
       }
-    }
 
-    for (const i of integrations) {
-      if (i.platform === PlatformType.GITHUB) {
-        for (const org of (i.settings as any).orgs) {
-          for (const repo of org.repos) {
-            const label = `${org.name}/${repo.name}`
-            const fullUrl = `https://github.com/${label}`
-            addToResult(i.platform, fullUrl, label)
+      const addToResult = (platform: PlatformType, fullUrl: string, label: string) => {
+        const platformKey = platform.toLowerCase()
+        if (!result[platformKey].some((item) => item.url === fullUrl)) {
+          result[platformKey].push({ url: fullUrl, label })
+        }
+      }
+
+      for (const i of integrations) {
+        if (i.platform === PlatformType.GITHUB) {
+          for (const org of (i.settings as any).orgs) {
+            for (const repo of org.repos) {
+              const label = `${org.name}/${repo.name}`
+              const fullUrl = `https://github.com/${label}`
+              addToResult(i.platform, fullUrl, label)
+            }
+          }
+        }
+
+        if (i.platform === PlatformType.GITHUB_NANGO) {
+          for (const org of (i.settings as any).orgs) {
+            for (const repo of org.repos) {
+              const label = `${org.name}/${repo.name}`
+              const fullUrl = `https://github.com/${label}`
+              addToResult(PlatformType.GITHUB, fullUrl, label)
+            }
+          }
+        }
+
+        if (i.platform === PlatformType.GIT) {
+          for (const r of (i.settings as any).remotes) {
+            let label = r
+            if (r.includes('https://gitlab.com/')) {
+              label = r.replace('https://gitlab.com/', '')
+            } else if (r.includes('https://github.com/')) {
+              label = r.replace('https://github.com/', '')
+            }
+            addToResult(i.platform, r, label)
+          }
+        }
+
+        if (i.platform === PlatformType.GITLAB) {
+          for (const group of Object.values((i.settings as any).groupProjects) as any[]) {
+            for (const r of group) {
+              const label = r.path_with_namespace
+              const fullUrl = `https://gitlab.com/${label}`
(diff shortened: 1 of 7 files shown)


backend/src/services/integrationService.ts
```diff
@@ -133,27 +135,88 @@ export default class IntegrationService {
...
+
+    const data: Partial<ICreateInsightsProject> = {}
+    const { platforms, widgets } = await collectionService.findSegmentsWidgetsById(segmentId)
+    data.widgets = widgets
+
+    if (platforms.some(IntegrationService.isCodePlatform)) {
+      const repositories = await collectionService.findRepositoriesForSegment(segmentId)
+      data.repositories = [
+        ...new Set(Object.values(repositories).flatMap((entries) => entries.map((e) => e.url))),
+      ]
```
joanagmaia: We should deduplicate repositories here, because repositories from git and github might be the same. As we usually track all repositories from both platforms

backend/src/services/integrationService.ts
```diff
@@ -133,27 +135,88 @@ export default class IntegrationService {
...
       })
 
+      const collectionService = new CollectionService(this.options)
+
+      const [insightsProject] = await collectionService.findInsightsProjectsBySegmentId(
+        record.segmentId,
+      )
+
+      if (insightsProject) {
+        this.updateInsightsProject(insightsProject.id, record.segmentId)
```
joanagmaia: If it's an "update integrations" we probably should have some sort of flag passed to "updateInsightsProject" from preventing to run `findGithubInsightsForSegment` every time we update an integration. We should only call this method if the github integration was created/updated.

backend/src/services/integrationService.ts
```diff
@@ -170,7 +233,21 @@ export default class IntegrationService {
...
+            if (integration?.segmentId) {
+              const [insightsProject] = await collectionService.findInsightsProjectsBySegmentId(
+                integration.segmentId,
+              )
+
+              if (insightsProject) {
+                await collectionService.updateInsightsProject(insightsProject.id, {
+                  repositories: [],
+                  widgets: [],
+                })
```
joanagmaia: We should add some validations here. We should only completely clean repositories and widgets if ALL segments integrations were destroyed.

Since this iterates through an array of destroyed integrations we should probably check by the end of this processing which platforms were destroyed, and for those we remove the repositories for the specific platforms, and the widgets for the specific platforms. Something like:
1. Delete integrations from the ids input
2. Check which integration platforms were deleted
3. For the repositories linked to the deleted platform, remove them from insightsProjects.repositoreis
4. For the widgets linked to the deleted platform, remove them from insightsProjects.widgets

joanagmaia (changes requested):

backend/src/services/integrationService.ts
ulemons: So, we don't want duplicates, right? Since we're initializing a new Set(), it should automatically prevent duplicates. Does that make sense?

joanagmaia: Oops, you're absolutely right.

backend/src/services/integrationService.ts
```diff
@@ -133,27 +135,88 @@ export default class IntegrationService {
...
+    data.widgets = widgets
+
+    if (platforms.some(IntegrationService.isCodePlatform)) {
+      const repositories = await collectionService.findRepositoriesForSegment(segmentId)
+      data.repositories = [
+        ...new Set(Object.values(repositories).flatMap((entries) => entries.map((e) => e.url))),
+      ]
+    }
+
+    if (platforms.includes(PlatformType.GITHUB)) {
```
joanagmaia: GITHUB or GITHUB_NANGO

backend/src/services/integrationService.ts
```diff
@@ -133,27 +136,114 @@ export default class IntegrationService {
...
+
+    const data: Partial<ICreateInsightsProject> = {}
+    const { widgets } = await collectionService.findSegmentsWidgetsById(segmentId)
+    data.widgets = widgets
+
+    if (IntegrationService.isCodePlatform(platform)) {
+      const repositories = await collectionService.findRepositoriesForSegment(segmentId)
+      data.repositories = [
+        ...new Set(Object.values(repositories).flatMap((entries) => entries.map((e) => e.url))),
+      ]
```
joanagmaia: I believe there is a problem with this one when we have a sub-project with repositories mapped to other sub-projects. We should exclude the repositories that are mapped to different sub-projects from this list.

backend/src/services/integrationService.ts
```diff
@@ -133,27 +136,114 @@ export default class IntegrationService {
...
+
+      if (insightsProject) {
+        await this.updateInsightsProject({
+          insightsProjectId: insightsProject.id,
+          isFirstUpdate: true,
+          platform: data.platform,
+          segmentId: record.segmentId,
+          transaction,
+        })
+      }
```
joanagmaia, line 170: For some reason, I don't think the `findGithubInsightsForSegment` is being triggered when isFirstUpdate is set to true. I created a github integration and it didn't update any of the properties.
And I know that there were availalbe properties because I tried to create the insights project manually which calls the endpoint to get github insights and it had results.

Probably worth debugging

backend/src/services/collectionService.ts
```diff
@@ -398,142 +396,133 @@ export class CollectionService extends LoggerBase {
...
 
-      if (i.platform === PlatformType.GITHUB_NANGO) {
-        for (const org of (i.settings as any).orgs) {
-          for (const repo of org.repos) {
-            const label = `${org.name}/${repo.name}`
-            const fullUrl = `https://github.com/${label}`
-            addToResult(PlatformType.GITHUB, fullUrl, label)
+      // Add mapped repositories to GitHub platform
+      const segmentRepository = new SegmentRepository(this.options)
+      const mappedRepos = await segmentRepository.getMappedRepos(segmentId)
```
joanagmaia, line 420: All the logic on this method is ok. But now that I look into `getMappedRepos`, I think we should simplify it because it won't return exactly what we want.

The "getMappedRepos" gets repositories that are assigned to a segment but whose associated integration belongs to a different segment (or no integration at all).

Meaning that it will work well for projects that are repos mapped from other integrations, but it won't work well for projects that have repos within its own integration.

I think if we change the query on `getMappedRepos`to:
```
select
         r.url as url
       from
        "githubRepos" r
       where r."segmentId" = :segmentId
       and r."tenantId" = :tenantId
       order by r.url
```
It should work

backend/src/services/collectionService.ts
```diff
@@ -398,142 +396,133 @@ export class CollectionService extends LoggerBase {
...
-      name,
-      topics,
-      twitter,
-      website,
-    }
+      return {
+        description,
+        github,
+        logoUrl,
+        name,
```
joanagmaia: Yesterday while I was using this feature in production. By creating projects manually, I found that using the `name` as the one that is returned on the github organization might not be ideal.
I think we should stick to the segment name.

Could we update this return payload so that:
```
name: segment.name
```
?

joanagmaia (approved):
LGTM!

PR #3241, opened 2025-07-08
chore: script to delete `activityRelations` with sourceId null

skwowet:


skwowet:
With recent improvements, we no longer need the script. Closing the PR.

PR #3247, opened 2025-07-10
chore: add script to fix-members-activities-after-unaffilation

mbani01:
# Changes proposed ✍️
Implement MemberUpdateBulk in order to trigger updateMember for multiple member in order to fix the activities having organizationID != null after unaffilation
### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/backend/src/bin/scripts/fix-members-activities-after-unaffilation.ts
+++ b/backend/src/bin/scripts/fix-members-activities-after-unaffilation.ts
@@ -1,0 +1,121 @@
+import commandLineArgs from 'command-line-args'
+import { randomUUID } from 'crypto'
+
+import { getDbConnection } from '@crowd/data-access-layer/src/database'
+import { getServiceLogger } from '@crowd/logging'
+import { getTemporalClient } from '@crowd/temporal'
+
+import { DB_CONFIG, TEMPORAL_CONFIG } from '@/conf'
+
+const log = getServiceLogger()
+
+const options = [
+  {
+    name: 'organizationId',
+    alias: 'o',
+    typeLabel: '{underline organizationId}',
+    type: String,
+    description: 'The organization ID to process members for.',
+  },
+  {
+    name: 'dryRun',
+    alias: 'd',
+    type: Boolean,
+    description: 'Run in dry-run mode (show what would be processed).',
+  },
+  {
+    name: 'help',
+    alias: 'h',
+    type: Boolean,
+    description: 'Print this usage guide.',
+  },
+]
+
+const parameters = commandLineArgs(options)
+
+setImmediate(async () => {
+  const organizationId = parameters.organizationId
+  const dryRun = parameters.dryRun ?? false
+
+  log.info({ organizationId, dryRun }, 'Running script with the following parameters!')
+
+  const db = await getDbConnection({
+    host: DB_CONFIG.readHost,
+    port: DB_CONFIG.port,
+    database: DB_CONFIG.database,
+    user: DB_CONFIG.username,
+    password: DB_CONFIG.password,
+  })
+  const temporal = await getTemporalClient(TEMPORAL_CONFIG)
+
+  try {
+    const memberIds = await db.any(
+      `
+      SELECT DISTINCT ar."memberId" as id
+      FROM "activityRelations" ar
+      JOIN "memberOrganizations" mo
+        ON ar."memberId" = mo."memberId"
+        AND ar."organizationId" = mo."organizationId"
+      WHERE ar."organizationId" = $1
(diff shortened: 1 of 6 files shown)


skwowet:
Can we keep the script inside `script-executor-worker`? Just to keep things clean and separate from services. If it’s like a one-time or short-run script, we can also add it in the `backend` service and run it like a regular node script.

mbani01:
@skwowet script moved to backend ✅

backend/src/bin/scripts/fix-members-activities-after-unaffilation.ts
```diff
@@ -0,0 +1,121 @@
...
+      SELECT DISTINCT ar."memberId" as id
+      FROM "activityRelations" ar
+      JOIN "memberOrganizations" mo
+        ON ar."memberId" = mo."memberId"
+        AND ar."organizationId" = mo."organizationId"
+      WHERE ar."organizationId" = $1
+        AND mo."deletedAt" IS NOT NULL;
+      `,
+      [organizationId],
+    )
```
joanagmaia: Hey, I think the query should be:
```
SELECT DISTINCT ar."memberId" AS id
FROM "activityRelations" ar
JOIN "memberOrganizations" mo
  ON ar."memberId" = mo."memberId"
  AND ar."organizationId" = mo."organizationId"
LEFT JOIN "memberOrganizationAffiliationOverrides" moao
  ON mo."id" = moao."memberOrganizationId"
WHERE ar."organizationId" = '3aa7e984-47e1-49c6-82ad-100fa4457604'
  AND (
    -- Case 1: The only membership for this org is deleted
    (
      mo."deletedAt" IS NOT NULL
      AND NOT EXISTS (
        SELECT 1
        FROM "memberOrganizations" mo2
        WHERE mo2."memberId" = mo."memberId"
          AND mo2."organizationId" = mo."organizationId"
          AND mo2."deletedAt" IS NULL
      )
    )
    -- Case 2: There’s an override that denies affiliation, and the membership is active
    OR (
      mo."deletedAt" IS NULL
      AND moao."allowAffiliation" = false
    )
  );
```

Case 1:
Only include members whose only affiliation entry for that organization is deleted. If a member has another (non-deleted) entry for the same organizationId in memberOrganizations, they should not be included in that part of the query.

Case 2:
There's an override in memberOrganizationAffiliationOverrides (MOAO) for the same member + organization, where:
allowAffiliation = false
And the associated memberOrganizations.deletedAt IS NULL (i.e., not deleted)

mbani01: Updated, thanks for the review!

joanagmaia (approved):
LGTM 🚀

services/libs/data-access-layer/src/old/apps/profiles_worker/index.ts
```diff
@@ -349,8 +349,11 @@ export async function runMemberAffiliationsUpdate(
     qx,
     memberId,
   )
+  logger.info(`orgCase for member ${memberId}: ${orgCases}`)
+  logger.info(`fullCase for member ${memberId}: ${fullCase}`)
```
skwowet, line 353: @mbani01 can we rm these debugger logs?

services/libs/data-access-layer/src/old/apps/profiles_worker/index.ts
```diff
@@ -349,8 +349,11 @@ export async function runMemberAffiliationsUpdate(
...
 
   const { minTimestamp, maxTimestamp } = await getMemberActivityTimestampRanges(qDb, memberId)
+  logger.info(`MemberActivityTimestampRanges ${minTimestamp} ${maxTimestamp}`)
```
skwowet, line 356: same here

skwowet (approved):
Looks good 👌🏻  

left a comment regarding the debugger logs in case you missed it.

skwowet:
@mbani01 just noticed there's a typo in the PR title 😄

services/libs/data-access-layer/src/old/apps/profiles_worker/index.ts
mbani01: Actually, the logs are added on purpose to ensure that the queries responsible for updating activities are generated correctly, since we couldn't identify why some activities were not updated as expected.

mbani01: Yeah, they can be spammy, but we can remove them once the script finishes, or have better visibility on the root cause

mbani01:
Thanks @skwowet 🙏

PR #3267, opened 2025-07-16
fix: fix a date type mismatch in the search volume pipe and data source

borfast:
The pipe was not returning any data due to a type mismatch in one of the date fields.


borfast:
> I also made a change in this datasource recently, when I found out that the datasource on the remote and the code were not the same, please see [this commit](https://github.com/CrowdDotDev/crowd.dev/pull/3284/commits/811748aaea5e4f56295c8b74a61b5af5ff6b441e)

Yeah, I noticed that, and frankly I forgot to revisit this PR of mine. I think this one can be closed since yours ended up taking care of the same thing.

PR #3271, opened 2025-07-17
fix: dedup activities array before upsert in data-sink-worker

skwowet:


services/apps/data_sink_worker/src/service/activity.service.ts
```diff
@@ -1174,7 +1174,31 @@ export default class ActivityService extends LoggerBase {
...
+      // only include non-null, non-empty values in the key
+      const key = [
+        item.payload.timestamp,
+        item.payload.platform,
+        item.payload.type,
+        item.payload.sourceId,
+        item.payload.channel,
+        item.payload.segmentId,
+      ]
+        .filter((v) => v !== undefined && v !== null && v !== '')
```
themarolt: To avoid any weird undefined behavior I would instead of ignoring empty values replace them with `<empty>` string or smt like that just to keep the format of a key consistent.

themarolt (changes requested):

services/apps/data_sink_worker/src/service/activity.service.ts
skwowet: Good catch - replaced empty values with <empty> for consistency.

PR #3282, opened 2025-07-23
Feature/prevent duplicates

ulemons:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


backend/src/database/migrations/V1753110221__create_insights_repo_table.sql
```diff
@@ -0,0 +1,107 @@
...
+    PRIMARY KEY (repository, "insightsProjectId")
+);
+
+-- 2. Enforce that a repository can be assigned to only one active project
+CREATE UNIQUE INDEX IF NOT EXISTS unique_active_insights_project_repo
+ON "insightsProjectsRepositories" (repository)
+WHERE "deletedAt" IS NULL;
+
+-- 3. Create or replace the sync function
+CREATE OR REPLACE FUNCTION sync_repositories()
```
themarolt: Hm so far we didn't use triggers or functions or procedures in the db itself so I'm wondering if it would be better to have this logic in the code itself?

ulemons: I was thinking of doing it this way for performance reasons. We can also implement it in code and see how it performs.

themarolt: Ye I would do that instead of triggers as it hides somewhat the logic. I mean if we were using this before I would let it be but if it's gonna be the only function we have in the db with such logic it will be hidden in a sense where people won't expect to find the logic here :)

ulemons: that makes perfectly sense, thanks!

themarolt (approved):

backend/src/database/migrations/V1753110221__create_segment_repos_table.sql
```diff
@@ -0,0 +1,17 @@
...
+    "archived" BOOLEAN NOT NULL DEFAULT FALSE,
+    "deletedAt" TIMESTAMPTZ,
+
+    PRIMARY KEY (repository, "segmentId")
+);
+
+-- 2. Enforce that a repository can be assigned to only one active segment
+CREATE UNIQUE INDEX IF NOT EXISTS unique_active_segment_repos
+ON "segmentRepositories" (repository)
+WHERE "deletedAt" IS NULL;
```
borfast: Shouldn't we add the unique index for every row, even if it's deleted, to avoid issues in case we ever remove the soft deleted marker?

borfast (approved):
Approved. 👍  Just one question that isn't critical.

backend/src/database/migrations/V1753110221__create_segment_repos_table.sql
ulemons: I added this constraint to keep track of deletions. If we use a unique index for every row, I guess we must use hard deletion. If this is the case I think we can go directly with hard deletion

borfast: We don't have to have another unique index for each row because we already have the primary key on the repository and segmentId, which already makes them unique. Or am I missing something?

PR #3305, opened 2025-08-05
fix: prevent deadlocks in member affiliation updates

skwowet:
# Changes proposed ✍️

Multiple `organizationUpdate` workflows running in parallel caused database deadlocks when updating the same `activityRelations` records. When a member worked at multiple organizations with overlapping dates, different workflows would try to update the same activity simultaneously:

```
Workflow A (Company A): "Activity X should belong to Company A"
Workflow B (Company B): "Activity X should belong to Company B"

Both workflows try to update Activity X at the same time → DEADLOCK → RETRY → DEADLOCK
```

This caused activities to keep switching between different organizations and triggered unnecessary replication to Sequin.

**Fix**: Changed from parallel to sequential processing of member affiliations. Now workflows process one affiliation at a time instead of all at once, eliminating race conditions and preventing deadlocks.

### DIFF
--- a/services/libs/data-access-layer/src/member-organization-affiliation/index.ts
+++ b/services/libs/data-access-layer/src/member-organization-affiliation/index.ts
@@ -403,14 +403,22 @@ export async function refreshMemberOrganizationAffiliations(qx: QueryExecutor, m
     return
   }
 
-  // process timeline in parallel
-  const results = await Promise.all([
-    ...affiliations.map((affiliation) => processAffiliationActivities(qx, memberId, affiliation)),
-    processFallbackActivities(qx, memberId, earliestStartDate, fallbackOrganizationId),
-  ])
+  let processed = 0
+
+  // Process affiliations one by one (sequentially) to avoid race conditions
+  for (const affiliation of affiliations) {
+    processed += await processAffiliationActivities(qx, memberId, affiliation)
+  }
+
+  // Then process fallback activities
+  processed += await processFallbackActivities(
+    qx,
+    memberId,
+    earliestStartDate,
+    fallbackOrganizationId,
+  )
 
   const duration = performance.now() - start
-  const processed = results.reduce((acc, processed) => acc + processed, 0)
 
   logger.info({ memberId }, `Refreshed ${processed} activities in ${duration}ms`)
 }


services/libs/data-access-layer/src/member-organization-affiliation/index.ts
```diff
@@ -4,29 +4,28 @@ import { getLongestDateRange } from '@crowd/common'
...
   const isDateInInterval = (date: Date, start: Date | null, end: Date | null) => {
     return (!start || date >= start) && (!end || date <= end)
   }
 
   const findOrgsWithRolesInDate = (
     date: Date,
-    memberOrganizations: MemberOrganizationWithOverrides[],
-  ): MemberOrganizationWithOverrides[] => {
-    const p = memberOrganizations.filter((row) => {
+    affiliations: Array<AffiliationItem>,
```
epipav: Let's use AffiliationItem[] instead to be consistent

services/libs/data-access-layer/src/member-organization-affiliation/index.ts
```diff
@@ -71,9 +76,16 @@ async function prepareMemberOrganizationAffiliationTimeline(
       }
 
       // 2. get the two orgs with the most members, and return the one with the most members if there's no draw
-      const sortedByMembers = orgs.sort((a, b) => b.memberCount - a.memberCount)
-      if (sortedByMembers[0].memberCount > sortedByMembers[1].memberCount) {
-        return sortedByMembers[0]
+      // only compare member orgs (manual affiliations don't have memberCount)
+      const memberOrgsOnly = orgs.filter(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (row: any) => !row.segmentId,
```
epipav: we can use AffiliationItem type here instead of any

services/libs/data-access-layer/src/member-organization-affiliation/index.ts
```diff
@@ -84,94 +96,123 @@ async function prepareMemberOrganizationAffiliationTimeline(
...
-      Math.min(...memberOrgsWithDates.map((row) => new Date(row.dateStart).getTime())),
+    manualAffiliations: IManualAffiliationData[],
+    fallbackOrganizationId: string | null,
+  ): TimelineItem[] => {
+    const allAffiliationsWithDates = [...memberOrganizations, ...manualAffiliations].filter(
+      (row) => !!row.dateStart,
     )
 
-    const now = new Date()
+    const earliestStartDate = allAffiliationsWithDates.length
```
epipav: let's use `allAffiliationsWithDates.length  > 0` for clarity

epipav (approved):

PR #3306, opened 2025-08-05
refactor: deprecate conversations on frontend [CM-2461]

ulemons:
# Changes Proposed ✍️

### What
We aim to deprecate the conversations feature. In this PR, we have removed all frontend components that interact with the conversations system.

🔗 closes: [CM-2461​](https://linear.app/lfx/issue/CM-2461/deprecate-conversations)

## Checklist ✅


frontend/src/modules/dashboard/components/dashboard-activities.vue
```diff
@@ -96,8 +88,6 @@ const mapData = (data: any[]) => data.map((item) => ({
   value: item.count,
 }));
 
-const tab = ref('recentConversations');
```
joanagmaia, line 99: Can you bring this line back and set it to `recentActivities`. We still need `tab` instance, and the default should be activities now

PR #3317, opened 2025-08-11
Feat/improve prevent duplication

ulemons:
# Changes proposed ✍️

### What
copilot:summary
​
copilot:poem

### How
copilot:walkthrough

## Checklist ✅


backend/src/services/integrationService.ts
```diff
@@ -155,84 +155,132 @@ export default class IntegrationService {
...
+      const [insightsProject] = await collectionService.findInsightsProjectsBySegmentId(
+        integration.segmentId,
+      )
+
+      if (!insightsProject) {
+        this.options.log.info(
+          `The segmentId: ${integration.segmentId} does not have any InsightsProject related`,
+        )
+        return integration
+      }
```
joanagmaia: I think this part:
```
if (IntegrationService.isCodePlatform(platform) && platform !== PlatformType.GIT) {
        this.gitConnectOrUpdate({
          remotes: repositories,
        })
      }
```
Should be before this code block.
Let's say that we don't have an insights project, we still want the git integration to be updated.
And the way that it's ordered, we wouldn't update the git integration (if I'm seeing this correctly).

Besides this I think everything else makes sense, because everything else is mainly based on the insightsProjectId.

services/libs/data-access-layer/src/collections/index.ts
```diff
@@ -326,61 +326,53 @@ export async function findBySlug(qx: QueryExecutor, slug: string) {
...
-      '("repository", "insightsProjectId") DO NOTHING',
-    ),
+    `
+    WITH "input" AS (
+      SELECT DISTINCT unnest(ARRAY[$(repositories:csv)]::text[]) AS "repository"
+    )
+    INSERT INTO "segmentRepositories" ("repository", "segmentId", "insightsProjectId")
+    SELECT "repository", $(segmentId), $(insightsProjectId)
+    FROM "input"
+    ON CONFLICT ("repository", "insightsProjectId")
```
joanagmaia: Should it be
ON CONFLICT ("repository") instead of ON CONFLICT ("repository", "insightsProjectId") .
Imagine a repository is already mapped in the table and you want to update it, the insightsProjectId would always be different, we only want to check the conflict on the repository side. So if we are trying to update the same repository, then both insightsProjectId and segmentId would be udpated to whatever we give it as input

joanagmaia (approved):

PR #3320, opened 2025-08-11
feat: add the archived repositories checker [IN-551]

borfast:
Implements the recurring archived repository checker, for [IN-551](https://linuxfoundation.atlassian.net/browse/IN-551).

Since this was asked to be a cronjob, it doesn't use the usual Temporal structure. It is meant to be executed periodically with a Kubernetes CronJob resource (added on the crowd-kube repository) but can also be executed independently, if necessary.

Instead of Temporal, it uses BullMQ to manage the task queue, which is much simpler and runs locally, needing only Redis.

This can also serve as an example / template for other simple recurring tasks we want to run which don't require the complexity of Temporal, and maybe allows us to slowly start moving away from it to reduce costs and simplify things.

### DIFF
--- a/services/cronjobs/archived_repositories/src/main.ts
+++ b/services/cronjobs/archived_repositories/src/main.ts
@@ -100,6 +100,10 @@ function prepareJobsByPlatform(repoURLs: string[]): { githubJobs: JobData[]; git
         url,
         platform: parsedResult.platform,
       },
+      opts: {
+        removeOnComplete: 1000,
+        removeOnFail: 5000
+      }
     };
 
     switch (parsedResult.platform) {


ulemons (approved):

services/cronjobs/archived_repositories/src/clients/github.ts
```diff
@@ -0,0 +1,10 @@
+import { ofetch } from 'ofetch';
+import { Config } from "../config";
+
+export async function isGitHubRepoArchived(owner: string, repo: string, config: Config): Promise<boolean> {
+  const data = await ofetch(`https://api.github.com/repos/${owner}/${repo}`, {
+    headers: { Authorization: `Bearer ${config.GithubToken}` },
+  });
+
```
mbani01, line 8: How are we handling unsuccessful responses here? I see there is an existing logic to limit the number of requests to avoid rate limits, but what will happen if we receive any other response besides 200?

borfast: Yesterday I asked Joana about the same thing, more specifically about when we get an error from GitHub (for example) saying the repository does not exist. We decided to add a new `deleted` column to track that but we didn't talk about the general error scenario.

BullMQ can [automatically retry failed jobs](https://docs.bullmq.io/guide/retrying-failing-jobs) depending on settings, so I would say we could use that as a first measure. After that, BullMQ puts all the failed jobs in a special `failed` queue, to which we can later add some monitoring if we want to.

mbani01: Alright, I see 👍 
My only concern was silent failures, but since we're tracking them, that should be fine

mbani01 (approved):

services/cronjobs/archived_repositories/src/clients/github.ts
borfast: Yeah, I agree. I also don't like silent failures but also thought it wasn't super critical because we are at least tracking the failures.

I'd still prefer to be warned, though, rather than only becoming aware of an issue after a user complains that their archived repository is still being used for metrics.

We can use BullMQ's events to warn us about failures by sending a Slack message or email, for example. I'll make a note of this to speak with Joana about it and implement it next.

borfast: Oh, and if we want to get fancy, we can also install a dashboard for BullMQ. There are a few of them, like [this](https://github.com/felixmosh/bull-board), [this](https://github.com/bee-queue/arena), and [this](https://taskforce.sh/).

mbani01: That would be great! Otherwise, if you're already storing them in the db, we can easily add a metaplane monitor since it's already configured to trigger Slack alerts (can be done in less than 5min), but not fancy 😅

borfast: Oh, that would be awesome! Let's talk about that in Slack. It seems like a very good idea and something we could do without much work. :+1:

services/cronjobs/archived_repositories/src/config.ts
```diff
@@ -0,0 +1,87 @@
+import dotenv from 'dotenv';
```
themarolt, line 1: Not really sure if having a separate .env file just for this service and using `dotenv` is the correct way now that all other services just use the `backend/.env...` files and the `backend-config` config maps on kubernetes clusters.

borfast: If you think it's better to have everything in the `backend-config` ConfigMap, I can change it.

The idea for the .env file is mainly for someone to be able to run the service locally. For production it is not meant to be used, as the shared configuration details (e.g. database credentials), are still loaded from the `backend-config` ConfigMap, and there would be a separate ConfigMap with the specific things that only this service uses (e.g. batch settings, or the GitHub and Gitlab tokens created just for this)

I figured I should not pollute the global ConfigMap for everyone byt adding the specific things there, if no other service will use them. Also, the principles of least privilege and data segregation would say that each service should only have access to the minimum amount of data and permissions, and putting everything in the same ConfigMap grants every service access to everything.

But I actually forgot to include this, so I was just working on that now. I pushed a couple of commits that clarify this in this repo, and another commit in the crowd-kube repo with a ConfigMap for this service.

That said, if you think it's better to put everything in the same ConfigMap, I can easily change it.

themarolt: I think it's gonna be clearer if everything is in one place as is the case now. cc @anilb0stanci what do you think?

borfast: OK, no worries, I'll merge them.

joanagmaia:
@borfast can you just confirm that all comments are tackled? For instance, the env.example file.
If we can also try to make sure there's no other issue with GitGuardian.

If so, I believe we can already merge this one!

borfast:
@joanagmaia, all the comments were taken care of, including the .env file. The .env.example file is still there to make things easier for local development with Docker Compose.

PR #3322, opened 2025-08-11
refactor: improve cross-identity member resolution for certain platforms in data-sink-worker

skwowet:
## Changes proposed ✍️

Added cross-type member resolution in `processActivities` for platforms that store email as username (Jira, Confluence, Gerrit). Git is excluded as we are not sure if the email is valid or not.

What happens now:

- Verified EMAIL identities search for verified USERNAME matches when usernames are email-like.
- Verified USERNAME identities (that look like emails) search for verified EMAIL matches.

This prevents duplicate members when the same value is stored as EMAIL in one platform and USERNAME in another, while preserving existing same-type merge logic.


services/apps/data_sink_worker/src/service/activity.service.ts
```diff
@@ -832,43 +846,195 @@ export default class ActivityService extends LoggerBase {
...
+            p.dbObjectMemberSource = 'email'
+          },
+        )
+      }
+
+      // Look up members using cross-identity matching (different platforms)
+      // we will check only on platforms that store email identities as usernames
+
+      // regex to detect if a username looks like an email address
+      const EMAIL_REGEX = /^[^\s@]+@[^\s@]+\.[^\s@]+$/i
```
themarolt: please add this to common lib or check if it's already there and make it an exported function so we can use it across the system

skwowet: Good point, let me check if it’s already there, or add this regex to the libs :v:

PR #3326, opened 2025-08-12
feat: data pipeline changes for segmentRepositories [IN-552] [IN-554]

borfast:
In the context of [IN-524](https://linuxfoundation.atlassian.net/browse/IN-524), we want to have the archived and excluded repositories for each project returned along with the project data.

This means adding the repository data to TinyBird, which already includes the archived and excluded flags in the `segmentRepositories` table, and then merge that data into the existing pipes that return project data, namely the `projects_list` and the `search_collections_projects_repos` pipes.

This PR:
* Adds the database migration to ensure we get data from the `segmentRepositories` table into Sequin.
* Adds a new `segmentRepositories` data source in TinyBird to receive data from the aforementioned table.
* Modifies the `insights_projects_populated_copy` pipe to include the list of archived and excluded repositories, which is then propagated to the `insights_projects_populated_ds` data source, from there it feeds the `insightsProjects_filtered` pipe, and finally the  `projects_list` pipe, which is the end goal.
  * The `search_collections_projects_repos` contains `archived` and `excluded` columns for the repositories results. This required adding a `JOIN` to the query, and with all the optimisations that were done not too long ago, I'm not sure if this 

Specific tickets in Jira:
* https://linuxfoundation.atlassian.net/browse/IN-552
* https://linuxfoundation.atlassian.net/browse/IN-554


backend/src/database/migrations/V1753798345__segmentRepositories_replication.sql
```diff
@@ -0,0 +1,3 @@
+ALTER PUBLICATION sequin_pub ADD TABLE "segmentRepositories";
+ALTER TABLE "segmentRepositories" REPLICA IDENTITY FULL;
+GRANT SELECT ON "segmentRepositories" to sequin;
```
epipav, line 3: we don't need this grant - sequin user only works on the sequin schema

services/libs/tinybird/datasources/insights_projects_populated_ds.datasource
```diff
@@ -9,6 +9,8 @@ DESCRIPTION >
     - `logoUrl`, `organizationId`, `website`, `github`, `linkedin`, `twitter` contain project branding and social links.
     - `widgets` array defines which insights widgets are enabled for this project.
     - `repositories` array contains the list of repository URLs associated with the project.
+    - `archivedRepositories` contains a list of archived repositories for the project.
+    - `excludedRepositories` contains a list of excluded repositories for the project.
```
epipav: would be nice to add short descriptions about what does it mean a repo is excluded or archived

services/libs/tinybird/datasources/segmentRepositories.datasource
```diff
@@ -0,0 +1,26 @@
...
+    - `last_archived_check` is the timestamp of the last check for whether the repository is archived.
+
+SCHEMA >
+    `repository` String `json:$.record.repository`,
+    `segmentId` UUID `json:$.record.segmentId`,
+    `insightsProjectId` UUID `json:$.record.insightsProjectId`,
+    `createdAt` DateTime64(3) `json:$.record.createdAt`,
+    `archived` Bool `json:$.record.archived`,
+    `excluded` Bool `json:$.record.excluded`,
+    `last_archived_check` Nullable(DateTime64(3)) `json:$.record.last_archived_check`
```
epipav: Instead of using Nullable fields, it's better to use Non-nullable fields with default values for performance reasons

services/libs/tinybird/datasources/segmentRepositories.datasource
```diff
@@ -0,0 +1,26 @@
...
+
+
+ENGINE ReplacingMergeTree
+ENGINE_PARTITION_KEY toYear(createdAt)
+ENGINE_SORTING_KEY repository
+ENGINE_VER createdAt
```
epipav: You'll need an `updatedAt` field to keep track of different versions of the same row. `createdAt` won't change on consecutive updates to the same row (when it's archived or excluded, for instance). So it can even pick the old row when an update happens

epipav (changes requested):

services/libs/tinybird/datasources/segmentRepositories.datasource
borfast: Oh, yeah, I forgot about that. The original table Umberto created doesn't have it and I wondered if it was necessary, but forgot to check it.

services/libs/tinybird/datasources/segmentRepositories.datasource
borfast: The column is a datetime and nullable on Postgres, and I think we might have had similar situations before - how are we converting null rows to non-null for Tinybird? Do we do a transformation in Sequin, for example?

backend/src/database/migrations/V1753798345__segmentRepositories_replication.sql
borfast: Doesn't it need access to do the backfills? I confess I never checked how it does them, but I assumed it was just querying the database normally.

services/libs/tinybird/datasources/insights_projects_populated_ds.datasource
borfast: Done.

services/libs/tinybird/datasources/segmentRepositories.datasource
borfast: I added the column and also updated the worker code to update its timestamp whenever it updates a repository archived / excluded status.

borfast:
Ping @epipav

joanagmaia:
@borfast can you confirm if the PR is up to date with the latest comments? And if so, can you gather all the questions you need to align with the team and perhaps go over them on a quick sync

backend/src/database/migrations/V1753798345__segmentRepositories_replication.sql
epipav: Actually, you are right - double-checking this, I see many tables (even `segmentRepositories` have it currently) have select grants for the sequin user. Also referred to [here](https://sequinstream.com/docs/how-to/setup-rds#:~:text=select%20permission%20on%20all%20the%20tables%20you%20want%20to%20connect%20to%20Sequin.) Resolving this one

services/libs/tinybird/datasources/segmentRepositories.datasource
epipav: You also need to change the `ENGINE_VER` to `updatedAt` so that this field is used for versioning

services/libs/tinybird/datasources/segmentRepositories.datasource
epipav: For strings etc, it's better to use a non-nullable field with default values so that tb uses less space and can query these fields faster

Tho this field is DateTime, it's actually better to use Nullable(DateTime64(3)) because it's not easy to use a default value here. Resolving this as well :+1:

services/libs/tinybird/datasources/segmentRepositories.datasource
borfast: Good catch, I thought I had changed it. It's changed now :+1:

PR #3347, opened 2025-08-22
feat: add bot detection in data ingestion

skwowet:
# Changes proposed ✍️

Added bot detection logic to the `data-sink-worker` to automatically identify bot accounts during member creation. The detection system uses a three-tier approach:

- Strong patterns for immediate bot confirmation, covering clear indicators such as `[bot]` notation, well-known bot services (e.g., dependabot, renovate, coderabbit), and other  patterns.
- Known bots list for specific accounts that cannot be consistently detected with regex alone.
- Common patterns for broader automation keywords and platform prefixes, which are used to flag potential bots for LLM validation.

Bot flags provided by source integrations always take precedence over our detection logic, and suspected bots are explicitly flagged for LLM validation. Existing isBot values are preserved during member updates to ensure we do not overwrite integration-provided information.

Also, `blacklistedDomains` was converted from an array to a Set to improve lookup performance (O(1) vs O(n)).


mbani01 (approved):

services/libs/common_services/src/services/bot.service.ts
```diff
@@ -0,0 +1,79 @@
+import { isKnownBot } from '@crowd/common'
+import { Logger, LoggerBase } from '@crowd/logging'
+import { IMemberIdentity, MemberBotDetection } from '@crowd/types'
+
+export class BotService extends LoggerBase {
```
epipav: NIT: Wouldn't it be better if we call this BotDetectionService instead?

services/apps/data_sink_worker/src/service/member.service.ts
```diff
@@ -98,6 +102,18 @@ export default class MemberService extends LoggerBase {
...
 
+          // detect if the member is a bot
+          const botDetection = this.botService.isMemberBot(data.identities, attributes)
+
+          if (botDetection === MemberBotDetection.CONFIRMED_BOT) {
+            this.log.debug({ memberIdentities: data.identities }, 'Member confirmed as bot.')
+
+            attributes.isBot = {
+              ...(attributes.isBot as object),
+              custom: true,
```
epipav: NIT: let's set `default: true` and instead of `custom` use another key to understand that this was actually set by the botDetectionService ?

epipav (approved):
Looks good! I added 2 nitpicks

services/libs/common_services/src/services/bot.service.ts
skwowet: Yep, makes sense

services/apps/data_sink_worker/src/service/member.service.ts
skwowet: Yeah, added a new `system` key for values set by BotDetectionService (or similar automated sources) and also, updated [memberAttValue](https://github.com/CrowdDotDev/crowd.dev/blob/improve/CM-521/services/apps/data_sink_worker/src/service/activity.service.ts#L1769) to respect this priority: 

```
custom > system > platform > default
```

PR #3349, opened 2025-08-26
feat: add fixtures for test tinybird pipes

ulemons:
# Changes proposed ✍️

### What
we want to test tinybird pipes locally.

### How
Added docs for testing tinybird locally. Also fixtures were added:
- `activityRelations_deduplicated_cleaned_ds.ndjson`
- `insightsProjects.ndjson`
- the `cli` uses the `TB_HOST` to connect to the local instance of `tinybird`
- `scaffold.insights.yaml` is forced to use the classic `tinybird` version
- in the `configure-tinybird-sink.sh` we are taking the token from the local `tinybird` instance
- the `tinybird-local-sink.properties` is using the `7181` instead of the `80`

-⚠️ remove fields from sequin.yml

## Checklist ✅

### DIFF
--- a/scripts/cli
+++ b/scripts/cli
@@ -681,14 +681,10 @@ function up_scaffold() {
     if [[ "$WITH_INSIGHTS" == "1" ]]; then
         download_kafka_connect_http "$CLI_HOME/scaffold/kafka-connect"
         install_tinybird_cli
-    fi
-
-    
-    if [[ "$WITH_INSIGHTS" == "1" ]]; then
         # first up tinybird, then wait for it to be ready, then set the token env for kafka connect
         $_DC --compatibility -p $PROJECT_NAME $DOCKER_COMPOSE_SCAFFOLD_FILES $DOCKER_COMPOSE_PROFILE up -d --build tinybird
-        fi
-    
+    fi
+
     $_DC --compatibility -p $PROJECT_NAME $DOCKER_COMPOSE_SCAFFOLD_FILES $DOCKER_COMPOSE_PROFILE up -d --build --no-recreate
 
 }


services/libs/tinybird/README.md
```diff
@@ -149,3 +149,81 @@ Until we move fully to **Tinybird Forward** (which will support migration script
...
+1. **Set Up Environment Variables and Authenticate**  
+   First, set up environment variables for both your local Tinybird instance and the staging environment, then authenticate using the Tinybird CLI. This will ensure the CLI is pointed to the correct environment when running commands. Replace `<YOUR_STAGING_TOKEN>` with your actual staging API token. The local admin token is retrieved via the local Tinybird API:
+   ```bash
+   # Set Tinybird local API endpoint and get the local admin token
+   export TB_HOST_LOCAL="http://localhost:7181"
+   TB_TOKEN_LOCAL="$(curl -s "$TB_HOST_LOCAL/tokens" | jq -r '.workspace_admin_token')"
+
+   # Set Tinybird staging API endpoint and token (replace with your staging info)
+   export TB_HOST_STAGING="https://api.us-west-2.aws.tinybird.co"   # use the correct region/domain for your case
+   export TB_TOKEN_STAGING="<YOUR_STAGING_TOKEN>"
```
borfast: I would suggest not telling people to run commands containing secret tokens in their shell because the token will most likely be recorded in the shell history.

Instead, perhaps we could tell people to have a `.env` file that they could `source` before running the `tb` command. The effect would be the same (env vars available in the shell) but the token would not be recorded in the shell history file.

Alternatively, though it seems too much for this, we could also suggest the use of [dotenvx](https://dotenvx.com/).

borfast (approved):
Good stuff! 👏 

I'm approving so this doesn't get blocked unnecessarily, but I'd love it if the instructions  were slightly different, as I mentioned in the comment.

epipav (approved):
lgtm :+1:

PR #3355, opened 2025-08-29
feat: software value service init

mbani01:
# Changes proposed ✍️
This pull request refactors the Go-based `software-value` service to simplify configuration management, streamline the Docker build process, and update how repositories are processed. The main changes include moving from TOML file-based configuration to environment variables, updating the Docker build to include Go-based binaries, and simplifying the service to process a single repository per run instead of batching multiple repositories.

- Added Go builder stage to build software-value binary and install scc to be copied to runner container
- Changing config init from file-based to env vars
- Processing a single repo instead of multiple ones
- Repo path is now passed as argument

### How
copilot:walkthrough

## Checklist ✅


scripts/services/docker/Dockerfile.git_integration
```diff
@@ -1,6 +1,27 @@
 # Base image for both stages
 FROM python:3.13.5-slim-bullseye AS base
 
+# Go builder stage: build the software-value binary and install scc
+FROM golang:1.24-bullseye AS go-builder
```
borfast: You can bump up Go from 1.24 to 1.25, and use a trixie base instead of bullseye for the most recent OS version (after Bullseye there was Bookworm, and now Trixie): `golang:1.25-trixie`

And you can even use the Alpine image, as it will be much smaller than the Debian one: `golang:1.25-alpine`

Debian-based: 845 MB
Alpine-based: 214 MB

It should make builds a little faster.

services/apps/git_integration/src/crowdgit/services/software_value/main.go
```diff
@@ -37,76 +42,31 @@ func main() {
...
-	if len(repoDirs) == 0 {
-		log.Printf("No git repositories found in %s", dirPath)
+	// Save to database
+	if err := insightsDb.saveProjectCost(ctx, report.Repository, report.Cocomo.CostInDollars); err != nil {
+		log.Fatalf("Error saving project cost: %v", err)
 	}
 
-	return repoDirs, nil
+	if err := insightsDb.saveLanguageStats(ctx, report.Repository, report.LanguageStats); err != nil {
+		log.Fatalf("Error saving language stats: %v", err)
```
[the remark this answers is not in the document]
borfast: I agree with Copilot, this should probably use log.Printf instead of log.Fatalf. Or did you have a specific reason to want to terminate this immediately if saving to the database fails?

mbani01:
@borfast Thanks for the review!
Suggestions applied, and included the git-integration/software-value communication.
As for the Koanf I agree it's more robust, I removed it since we're only getting the db config and thought maybe we can remove an extra dependency

services/apps/git_integration/src/crowdgit/services/software_value/software_value_service.py
```diff
@@ -1,10 +1,25 @@
...
+    async def run(self, repo_path: str):
+        """
+        Triggers software value binary for given repo.
+        Results are saved into insights database directly
+        """
+        try:
+            self.logger.info("Running software value...")
+            output = await run_shell_command([self.software_value_executable, repo_path])
+            self.logger.info(f"Software value output: {output}")
+            # json_output = json.loads(output)
```
mbani01: I'm adding the error handling in next PR for all services not only softare value, that's why this is commented

borfast:
All good. No arguments against removing a dependency. :+1: :slightly_smiling_face:

PR #3358, opened 2025-09-02
feat: enable git integration v2 for newly connected + updated repos

mbani01:
# Changes proposed ✍️
This pull request introduces support for syncing repositories to the new `git.repositories` table as part of the migration to git-integration V2. It adds a new repository class for managing the table, updates the integration service to handle upserts and soft deletes, and implements logic to inherit repository IDs from `githubRepos` for smoother migration.

**Git-integration V2 migration:**

* Added new `GitReposRepository` class with methods for soft deleting and upserting repositories in the `git.repositories` table.
* Updated `IntegrationService` to soft delete repositories in `git.repositories` when an integration is deleted.
* Implemented `syncRepositoriesToGitV2` method in `IntegrationService` to upsert repositories to `git.repositories`, inheriting IDs from `githubRepos` when migrating. [[1]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR1264-R1335) [[2]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR1230-R1254)

**Dependency and import updates:**

* Added necessary imports for `GitReposRepository` and `QueryTypes` in `integrationService.ts`. [[1]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdL7-R7) [[2]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR50)

### How
copilot:walkthrough

## Checklist ✅

### DIFF
--- a/backend/src/services/integrationService.ts
+++ b/backend/src/services/integrationService.ts
@@ -412,10 +412,27 @@ export default class IntegrationService {
               })
 
               // Also soft delete from git.repositories for git-integration V2
-              await GitReposRepository.delete(integration.id, {
-                ...this.options,
-                transaction,
-              })
+              try {
+                // Find the Git integration ID for this segment
+                const gitIntegration = await IntegrationRepository.findByPlatform(
+                  PlatformType.GIT,
+                  {
+                    ...this.options,
+                    currentSegments: [{ id: integration.segmentId } as any],
+                    transaction,
+                  },
+                )
+                if (gitIntegration) {
+                  await GitReposRepository.delete(gitIntegration.id, {
+                    ...this.options,
+                    transaction,
+                  })
+                }
+              } catch (err) {
+                this.options.log.info(
+                  'No Git integration found for segment, skipping git.repositories cleanup',
+                )
+              }
             }
           }
 


backend/src/services/integrationService.ts
```diff
@@ -1242,6 +1261,78 @@ export default class IntegrationService {
...
+      }))
+
+      if (repositoriesToSync.length === 0) {
+        this.options.log.warn('No githubRepos found - skipping repository sync to git v2')
+        return
+      }
+    } else {
+      // Generate new entries with auto-generated UUIDs
+      repositoriesToSync = remotes.map((url) => ({
+        id: require('uuid').v4(), // Generate new UUID
```
ulemons: Why are we using `require` here ? can we use a classic "improt { v4 as uuidv4} from 'uuid' ?

backend/src/services/integrationService.ts
```diff
@@ -1242,6 +1261,78 @@ export default class IntegrationService {
...
+          type: QueryTypes.SELECT,
+          transaction,
+        },
+      )
+
+      repositoriesToSync = (githubRepos as any[]).map((repo) => ({
+        id: repo.id,
+        url: repo.url,
+        integrationId,
+        segmentId: options.currentSegments[0].id,
```
ulemons, line 1340: is it safe to access the currentSegments in this way ? is this always going to be here ?

backend/src/database/repositories/gitReposRepository.ts
```diff
@@ -0,0 +1,81 @@
...
+    }>,
+    options: IRepositoryOptions,
+  ) {
+    const seq = SequelizeRepository.getSequelize(options)
+    const transaction = SequelizeRepository.getTransaction(options)
+
+    // Build SQL placeholders and parameter replacements in a single loop
+    const placeholders: string[] = []
+    const replacements: Record<string, any> = {}
+
```
ulemons, line 56: what if we have empty repositories here ? maybe we can have an early return ?

backend/src/services/integrationService.ts
```diff
@@ -1242,6 +1261,78 @@ export default class IntegrationService {
...
+   * @param integrationId The integration ID from the git integration
+   * @param inheritFromGithubRepos If true, queries githubRepos for IDs; if false, generates new UUIDs
+   */
+  private async syncRepositoriesToGitV2(
+    remotes: string[],
+    options: IRepositoryOptions,
+    transaction: Transaction,
+    integrationId: string,
+    inheritFromGithubRepos: boolean,
+  ) {
```
ulemons, line 1308: Usually, when a function has more than three parameters, I prefer using an object as input. It’s less error-prone since the order of the arguments doesn’t matter.

backend/src/services/integrationService.ts
mbani01: Good catch! I forgot it

backend/src/services/integrationService.ts
mbani01: I assume, because it's being used like this in many places in the code base 🤔

backend/src/database/repositories/gitReposRepository.ts
mbani01: Yes, it makes sense 👍

backend/src/services/integrationService.ts
mbani01: I agree, once the migration’s done this function will be simplified, adding a TODO so I don’t forget.

PR #3364, opened 2025-09-03
feat: add missing Gitlab and Gerrit activities in some pipes

borfast:
[Ticket in Jira](https://linuxfoundation.atlassian.net/browse/IN-686)

This adds the necessary changes to take activities from Gitlab and Gerrit into account in some of the pipes that feed the UI widgets.

### DIFF
--- a/services/libs/tinybird/pipes/active_days.pipe
+++ b/services/libs/tinybird/pipes/active_days.pipe
@@ -73,7 +73,15 @@ SQL >
                 'discussion-comment',
                 'merge_request-opened',
                 'merge_request-comment',
-                'merge_request-review-approved'
+                'merge_request-review-approved',
+                'merge_request-review-changes-requested',
+                'merge_request-assigned',
+                'merge_request-review-requested',
+                'merge_request-merged',
+                'changeset-closed',
+                'changeset_comment-created',
+                'patchset_comment-created',
+                'patchset_approval-created'
             )
         GROUP BY ds."startDate", ds."endDate"
         order by ds."startDate"
--- a/services/libs/tinybird/pipes/pull_request_analysis_copy_pipe.pipe
+++ b/services/libs/tinybird/pipes/pull_request_analysis_copy_pipe.pipe
@@ -13,13 +13,13 @@ SQL >
         memberId,
         organizationId
     FROM activityRelations_deduplicated_cleaned_ds
-    WHERE type = 'pull_request-opened'
+    WHERE type = 'pull_request-opened' OR type = 'merge_request-opened' OR type = 'changeset-created'
 
 NODE pull_request_first_assigned
 SQL >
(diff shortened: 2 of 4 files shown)


services/libs/tinybird/pipes/active_days.pipe
```diff
@@ -73,7 +73,15 @@ SQL >
...
-                'merge_request-review-approved'
+                'merge_request-review-approved',
+                'merge_request-review-changes-requested',
+                'merge_request-assigned',
+                'merge_request-review-requested',
+                'merge_request-merged',
+                'changeset-closed',
+                'changeset_comment-created',
+                'patchset_comment-created',
+                'patchset_approval-created'
```
joanagmaia: I believe the `patchset-created` is missing.
And that we should not include `merge_request-assigned` or `merge_request-review-requested` since they are not marked as contributions.

services/libs/tinybird/pipes/pull_request_analysis_copy_pipe.pipe
```diff
@@ -13,42 +13,53 @@ SQL >
...
+    WHERE type = 'pull_request-assigned' OR type = 'merge_request-assigned'
     GROUP BY sourceParentId
     order by min(timestamp) desc
 
 NODE pull_request_first_review_requested
 SQL >
     SELECT sourceParentId, MIN(timestamp) AS reviewRequestedAt
     FROM activityRelations_deduplicated_cleaned_ds
-    WHERE type = 'pull_request-review-requested'
+    WHERE type = 'pull_request-review-requested' OR type = 'merge_request-review-requested'
```
joanagmaia, line 30: Follow up question on this one. Since for gerrit we won't have these 2 metrics, how do we handle these cases? Would it mean that we would never show Gerrit metrics for the "Merge lead time" widget?

services/libs/tinybird/pipes/active_days.pipe
borfast: > I believe the patchset-created is missing.

Good catch. It's not listed in the ticket but it is part of the types definition in the code. Thanks :+1: 


> And that we should not include merge_request-assigned or merge_request-review-requested since they are not marked as contributions.

Well spotted. I think I mixed a few of these since the concepts are similar.

services/libs/tinybird/pipes/pull_request_analysis_copy_pipe.pipe
borfast: That's a good question. Since we don't have those metrics, Gerrit repos would never show this data for that widget, and it would just display the typical message saying there's no data.

joanagmaia: Let's not add anything for now, and see how it affects the production data for the projects that have Gerrit.

PR #3368, opened 2025-09-05
feat: add pipe for fetching the activities

ulemons:
# Changes proposed ✍️

### What
Added a new Tinybird pipe (`activities_relations_filtered.pipe`) with two nodes:
- `filtered_relations`: filters and paginates activity relations.
- `activities_enriched_v1`: enriches results with `url`, `body`, and `title` from activities.

### Why
We need a consistent, performant way to return relation metadata together with activity content, without scanning full datasets.

### How
The enrichment uses a join restricted to the small, paginated subset of `activityId`s from the first node, so performance scales with `pageSize` rather than dataset size.

## Checklist ✅


borfast:
Should we have CDP touching Tinybird? Not too long ago we decided that Insights should not touch the CDP database directly, and that each project should be isolated in terms of data storage access. Would there be another way to do this?

ulemons:
@borfast Sorry, I don’t think I fully understood what you mean, do you mean that Tinybird is considered the exclusive data source for insights? If that’s the case, then every time the CDP needs data from Tinybird insights it would have to expose APIs to provide those data, and I’m not sure that would be very convenient.

borfast:
@ulemons, yes, that's it. I know it's less convenient but having cross-service data sources is usually A Very Bad Idea™ :sweat_smile: 

I'll bring this up for discussion in Slack, as it's probably easier to talk about it there.

borfast:
@ulemons, nevermind, let's move forward with this. I was speaking with Joana about something else, ended up asking her about the inter-service DB dependency, and she explained it, so ignore my previous comments :+1:

themarolt (approved):

PR #3383, opened 2025-09-10
fix: persist primary organization even if role changes in affiliations

ulemons:
# Changes proposed ✍️

### What
When we have multiple affiliations for the same org, we want to persist the record of the `PrimaryOrg`.

## Checklist ✅


skwowet (changes requested):
LGTM. Can we remove all the debugger logs and add a good PR title?

skwowet (approved):

PR #3384, opened 2025-09-10
fix: sync GitHub disconnection to git integration [CM-675]

mbani01:
# Changes proposed ✍️
This pull request updates integration removal logic and environment configuration to better support the LFX Enterprise Edition and improve the handling of Git integrations. The main changes include switching the default edition to `lfx-ee`, enhancing the cleanup process for Git integrations, and improving robustness for segment-based operations.

**Edition and environment configuration:**
* Changed the default `CROWD_EDITION` in `.env.dist.local` from `community` to `lfx-ee`, aligning the configuration with LFX Enterprise Edition requirements.

**Integration removal and Git integration handling:**
* Refactored the integration removal process in `IntegrationService` to always use the LFX logic, removing the previous edition check and ensuring consistent behavior.
* Enhanced the logic for removing Git integrations: if all remotes are removed, the Git integration and its repositories are now fully deleted; otherwise, only the remaining remotes are updated.
* Improved the soft deletion of Git repositories by ensuring the correct segment and transaction context are used when finding and deleting Git integrations.

**Robustness improvements for segment-based operations:**
* Added null checks for `segmentId` before performing operations related to insights projects and widgets, preventing potential errors when `segmentId` is undefined.
* Ensured that segment repositories are only deleted if a valid `segmentId` is present, increasing code safety.

## Checklist ✅

### DIFF
--- a/backend/src/services/integrationService.ts
+++ b/backend/src/services/integrationService.ts
@@ -482,27 +482,20 @@ export default class IntegrationService {
       const qx = SequelizeRepository.getQueryExecutor(this.options)
 
       let insightsProject = null
+      let widgets = []
+
       if (segmentId) {
         const [project] = await collectionService.findInsightsProjectsBySegmentId(segmentId)
         insightsProject = project
-      }
-
-      let widgets = []
-      if (segmentId) {
         const widgetsResult = await collectionService.findSegmentsWidgetsById(segmentId)
         widgets = widgetsResult.widgets
-      }
-
-      const insightsRepo = insightsProject?.repositories ?? []
-
-      if (segmentId) {
         await deleteSegmentRepositories(qx, {
           segmentId,
         })
       }
 
+      const insightsRepo = insightsProject?.repositories ?? []
       const filteredRepos = insightsRepo.filter((repo) => !toRemoveRepo.has(repo))
-
       // remove duplicates
       const repositories = [...new Set<string>(filteredRepos)]
 


backend/.env.dist.local
```diff
@@ -1,6 +1,6 @@
 # Global settings
 KUBE_MODE=1
-CROWD_EDITION=community
```
mbani01, line 3: This change is out of scope for this PR, but the default should be lfx to avoid inconsistent local/prod behavior

backend/src/services/integrationService.ts
```diff
@@ -336,149 +336,170 @@ export default class IntegrationService {
...
       const insightsRepo = insightsProject?.repositories ?? []
 
-      await deleteSegmentRepositories(qx, {
-        segmentId,
-      })
+      if (segmentId) {
+        await deleteSegmentRepositories(qx, {
+          segmentId,
+        })
+      }
```
mbani01: @ulemons  For some reason, this fails locally due to missing `segmentId`. I didn't investigate further, but the check seems to work. And from a quick test this isn't happening on staging/prod so I assume it has something to do with local setup

ulemons: i ll do a quick test in local env. btw do you make sense to do everything inside one if (segmentId) block ? maybe it is easier to read. what do you think ?

ulemons: approved, for me is working locally 👍

mbani01: Definitely makes sense 👍 
It's fixed

PR #3386, opened 2025-09-10
feat(IN-529): add DB migrations and Tinybird data source for activity_types

borfast:
This PR introduces a new `activity_types` database table in CDP which will be the single source of truth for all the activity types we have.

It also introduces the equivalent Tinybird data source, to be populated via Sequin.

The migration file that creates the table also inserts all the necessary activity types, as described in the [PRD](https://docs.google.com/document/d/15en2exL4YmSWLW2UMwlAh-EPwu3qSjMtFixrHrVYkC8/edit?tab=t.0#heading=h.q1qhx9jzveqk).


backend/src/database/migrations/V1757490935__create_activity_types_table.sql
```diff
@@ -0,0 +1,116 @@
+CREATE TABLE activity_types (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    activity_type VARCHAR(150),
+    platform VARCHAR(150),
+    is_code_contribution BOOLEAN,
+    is_collaboration BOOLEAN,
```
gaspergrom: please stick to similar naming convention we have for other tables and make it activityTypes instead of activity_types

backend/src/database/migrations/V1757490935__create_activity_types_table.sql
```diff
@@ -0,0 +1,116 @@
...
+('thread_message', 'discord', false, true);
+
+-- Slack platform (Collaboration)
+INSERT INTO activity_types (activity_type, platform, is_code_contribution, is_collaboration) VALUES
+('message', 'slack', false, true);
+
+-- Stack Overflow platform (Collaboration)
+INSERT INTO activity_types (activity_type, platform, is_code_contribution, is_collaboration) VALUES
+('question', 'stackoverflow', false, true),
+('answer', 'stackoverflow', false, true);
```
gaspergrom: You can also include already migrations to add this table to the sequin replication slot

borfast: Good point, I'll add it :+1:

backend/src/database/migrations/V1757490935__create_activity_types_table.sql
borfast: Oh, I wanted to have brought that up and forgot about it.

We have a mix of snake_case and camelCase in some table names, so I wanted to propose that we standardise on the snake_case because it's what the majority of the world uses as the default, and avoids us having to quote table names.

But then I realised that this mix is actually happening on TinyBird and not on PostgreSQL, so it makes more sense to keep it camelCase in Postgres.

As for Tinybird, I guess for the data source name it makes sense to use the same casing as the original table. But for the pipes I would like to understand if we're following any pattern (e.g. dataSource_other_stuff), and if not, to try to standardise it. But that's something for another time.

backend/src/database/migrations/V1757490935__create_activityTypes_table.sql
```diff
@@ -0,0 +1,120 @@
...
+('issues-closed', 'github', false, true),
+('issue-comment', 'github', false, true),
+('discussion-comment', 'github', false, true);
+
+-- GitLab platform (Collaboration)
+INSERT INTO "activityTypes" ("activityType", platform, "isCodeContribution", "isCollaboration") VALUES
+('issues-opened', 'gitlab', false, true),
+('issues-closed', 'gitlab', false, true),
+('issue-comment', 'gitlab', false, true),
+('authored-commit', 'gitlab', false, true);
```
borfast: `authored-commit` is categorised as a code contribution for Git and GitHub but as a collaboration for GitLab. It doesn't seem right. Let's double check this before merging.

backend/src/database/migrations/V1757490935__create_activityTypes_table.sql
```diff
@@ -0,0 +1,120 @@
...
+    "activityType" VARCHAR(150),
+    platform VARCHAR(150),
+    "isCodeContribution" BOOLEAN,
+    "isCollaboration" BOOLEAN,
+    "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE INDEX "idx_activityTypes_activityType"
+    ON "activityTypes" ("activityType");
```
gaspergrom: On this index would be good just to have id and updated at in index as this is what sequin will use to distinguish what changed... on tinybird then have sorting by activity types so the datasource is sorted by key which is filtered the most

gaspergrom (changes requested):

backend/src/database/migrations/V1757490935__create_activityTypes_table.sql
joanagmaia: @borfast Jonathan confirmed in the document, authored-commit from gitlab should also be included as a codeContribution

backend/src/database/migrations/V1757490935__create_activityTypes_table.sql
borfast: I think on Tinybird the filtering will mostly be by `isCodeContribution` and `isCollaboration`. We will want to do sub-queries like 
```
SELECT *
FROM somewhere
WHERE activityType IN (SELECT activityType FROM activityTypes WHERE isCodeContribution = true AND platform = 'github')
```
But at the scale of this table it's hardly going to be noticeable, as we will have less than 70 rows.

In fact, I removed the partitioning because it's detrimental to performance on such a small table.

For the sorting, I can add something like `ENGINE_SORTING_KEY (isCodeContribution, isCollaboration, platform)`


On the PostgreSQL and Sequin side, it's a good point. I'll add an index on `updatedAd` for the backfills from Sequin and remove the index on `activityType` (I added it thinking we would filter by it, but not only will we not, as filtering will actually happen on Tinybird :man_facepalming: )

gaspergrom (approved):

PR #3404, opened 2025-09-16
chore: update security schemas

joanagmaia:
# Changes proposed ✍️

### What
- Update security schemas to accomodate new recommendation column and the new schema from the pvtr plugin according to layer 4 of gemara
- Fix privateer plugin to specific commit so that we can run or re-deploy the security-best-practices-worker always with the same version unless we want to update it.

## Checklist ✅

### DIFF
--- a/services/libs/tinybird/datasources/securityInsightsEvaluationAssessments.datasource
+++ b/services/libs/tinybird/datasources/securityInsightsEvaluationAssessments.datasource
@@ -36,7 +36,7 @@ SCHEMA >
     `message` String `json:$.record.message` DEFAULT '',
     `stepsExecuted` UInt8 `json:$.record.stepsExecuted` DEFAULT 0,
     `runDuration` String `json:$.record.runDuration` DEFAULT '',
-    `recommendation` String `json:$.record.recommendation` DEFAULT '',
+    `recommendation` Nullable(String) `json:$.record.recommendation`,
     `start` String `json:$.record.start` DEFAULT '',
     `end` Nullable(String) `json:$.record.end`,
     `value` Nullable(String) `json:$.record.value`,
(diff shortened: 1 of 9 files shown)


themarolt (approved):

gaspergrom (approved):

services/libs/tinybird/datasources/securityInsightsEvaluationAssessments.datasource
```diff
@@ -31,6 +36,11 @@ SCHEMA >
     `message` String `json:$.record.message` DEFAULT '',
     `stepsExecuted` UInt8 `json:$.record.stepsExecuted` DEFAULT 0,
     `runDuration` String `json:$.record.runDuration` DEFAULT '',
+    `recommendation` String `json:$.record.recommendation` DEFAULT '',
+    `start` String `json:$.record.start` DEFAULT '',
+    `end` String `json:$.record.end` DEFAULT '',
+    `value` String `json:$.record.value` DEFAULT '',
```
borfast: There's a type mismatch here for the default value: null in PostgreSQL, and an empty string in Tinybird. It will work, as Tinybird will convert a null value into an empty string, but this changes the meaning of the data.

In Postgres we can have either a JSON object, or a NULL value. The JSON object represents an actual value - the data exists, even if it's empty (`{}`). The NULL represents a missing or unknown value - there is no data.

We could improve this by making these columns `Nullable(String)` in Tinybird and defaulting to NULL instead of an empty string, to match what we have in PostgreSQL.

joanagmaia: That's a very good point, no reason.
Let me double check the types from Gemara layer 4 and then confirm both on postgres and tinybird to make sure they have the same types.
Thanks for flagging this!

joanagmaia:
Hey @borfast, updated the types to your suggestion.
`end`, `value`, and `changes` now have `Nullable(String)` so that default is null and not `''`

Can you double check?

borfast (approved):
👍 
I meant to have approved yesterday and have the previous comment only as a suggestion, sorry.

PR #3409, opened 2025-09-17
fix: safely decode shell command output

mbani01:
# Changes proposed ✍️
This pull request improves the robustness of shell command output handling in the git integration service by introducing safer decoding for potentially non-UTF-8 encoded data. The main change is the addition of a utility function to handle various encodings, ensuring that git command outputs are properly processed regardless of their original encoding.

**Shell command output decoding improvements:**

* Added a new `_safe_decode` function to `services/apps/git_integration/src/crowdgit/services/utils.py` that attempts to decode bytes using UTF-8, falls back to other common encodings, and finally replaces invalid bytes if necessary.
* Updated the `run_shell_command` function to use `_safe_decode` for both `stdout` and `stderr`, improving reliability when handling git command outputs with unexpected encodings.

## Checklist ✅

### DIFF
--- a/services/apps/git_integration/src/crowdgit/services/utils.py
+++ b/services/apps/git_integration/src/crowdgit/services/utils.py
@@ -13,6 +13,40 @@ from crowdgit.errors import (
 from crowdgit.logger import logger
 
 
+def _safe_decode(data: bytes) -> str:
+    """
+    Safely decode bytes to string, handling various encodings that might be present in git output.
+
+    Git repositories can contain commit messages and other text in various encodings.
+    This function attempts to decode using UTF-8 first, then falls back to other common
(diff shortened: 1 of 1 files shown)


services/apps/git_integration/src/crowdgit/services/utils.py
```diff
@@ -13,6 +13,40 @@
...
+    if not data:
+        return ""
+
+    try:
+        return data.decode("utf-8")
+    except UnicodeDecodeError:
+        pass
+
+    # Try common legacy encodings
+    for encoding in ["latin-1", "cp1252", "iso-8859-1"]:
```
skwowet: nitpick: can we use tuple?

I believe it is created once at import time and signals that the collection is constant.

PR #3437, opened 2025-09-30
feat: add platforms_by_segment.pipe

ulemons:
# Changes proposed ✍️

### What
For QuestDB deprecation, we need a query that returns the distinct platforms used by a set of segments, based on their activities.

### How
A new pipe is added: `pipes/platforms_by_segmtne.pipe`

## Checklist ✅

### DIFF
--- a/services/libs/tinybird/pipes/platforms_by_segments.pipe
+++ b/services/libs/tinybird/pipes/platforms_by_segments.pipe
@@ -1,0 +1,25 @@
+DESCRIPTION >
+  - Returns a single array of distinct platforms.
+  - Source: activityRelations_deduplicated_ds.
+  - Filters by segmentId (required at runtime) and optional time range (after/before).
+  - Notes:
+    - CamelCase identifiers are quoted.
+    - The query compiles without params. If SEGMENTS is missing, it returns empty results.
+
+NODE platforms_array
+SQL >
+  %
+  /* Distinct platforms as one array (no sort for max perf) */
+  SELECT
+      groupUniqArray("platform") AS platforms
+  FROM activityRelations_deduplicated_ds
+  WHERE
+      "segmentId" IN {{ Array(segments, 'String', required=True, description="Segment IDs") }}
+      {% if defined(after) and after %}
+        AND "timestamp" > parseDateTimeBestEffort({{ String(after) }})
+      {% end %}
+      {% if defined(before) and before %}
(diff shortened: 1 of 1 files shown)


borfast (commented):
Everything looks good but I'm not sure if we should use the `platform` column from `activityRelations_deduplicated_ds`, as it might be removed in the aftermath of [IN-529](https://linuxfoundation.atlassian.net/browse/IN-529) (PRD [here](https://docs.google.com/document/d/15en2exL4YmSWLW2UMwlAh-EPwu3qSjMtFixrHrVYkC8/edit?pli=1&tab=t.0)).

In that one, we added a new `activityTypes` table and pipe, and we want to remove the `isContribution` column from the `activityRelations` table and pipe (these changes would cascade into `activityRelations_deduplicated_ds`).

That alone would not affect your PR here, but it got me thinking that since we also have the `platform` in the new `activityTypes` table, we probably also want to remove it from `activityRelations`.

@joanagmaia, what do you say?

joanagmaia:
@borfast I might be wrong but I don't think we can remove platform from activityRelations or if we would want that.
Main reasons would be:
- For each activity we need to identify the activity type + platform
- Activity type is not unique, meaning that we might have the same "authored-commit" type for different platforms (e.g. git, github). So we would always require a combination of both to identify what the activity is referring to.
- I think the `isContribution` flag is different from platform in the sense that it's inferred based on a combination of both type and platform. So being able to remove `isContribution` from `activityRelations` doesn't necessarily mean we could remove `platform` or `type`.

And from the PR goal, it seems that this is a pipe that will be able to tell us "in which platforms is a project active in", so using activityRelations and platform seems like the correct approach for it 🤔 

I might be wrong, but yeah I would say that this is correct ✅

borfast:
@joanagmaia, got it, makes sense. But then that raises another point: we have two `platform` columns, one in `activityRelations` and another in `activityTypes`. But that's another topic.

If we're not removing `platform` from `activityRelations`, then this looks good to go :+1:

ulemons:
@joanagmaia @borfast thanks for the review guys, sorry I forgot to add that this PR has a dependency, meaning that if the assumption made on the this other [PR](https://github.com/CrowdDotDev/crowd.dev/pull/3439) is true then probably we can just close this PR since this is not going to be needed for the scope of the `QuestDB` deprecation.

ulemons:
as mentioned [here](https://github.com/CrowdDotDev/crowd.dev/pull/3439) I close this pr since is not needed anymore

PR #3439, opened 2025-09-30
refactor: remove findNewActivityPlatforms

ulemons:
# Changes proposed ✍️

### What
We are removing `findNewActivityPlatforms`. The idea behind this decision comes from the fact that in the `refreshDashboardCache` function, it seems impossible for `segmentsEnabled` to be false.

You can also double check this from the function `spawnDashboardCacheRefresh`, It seems that segmentId sent to the lower function can never be undefined.

As a result, the else branch is never actually executed, which means we can safely remove quite a bit of apparently unused code.

## Checklist ✅


themarolt (approved):

epipav (approved):
lgtm

PR #3432, opened 2025-09-29
feat: allow filtering contributions and collaborations [IN-707]

borfast:
This replaces the `isContribution` filter in the pipes that use activities, and adds the ability to filter by activity type "category", by including a new `includeCodeContributions` and `includeCollaborations` parameters.

By default, if neither parameter is passed to the pipe, it considers `includeCodeContributions = 1`.

The changes in `services/libs/tinybird/pipes/top_member_org_copy.pipe` were done so that the "Top contributors" and "Top organizations" widgets in the app front page use filtered data.

I'm not sure what to do for the "Top LF projects" widget, though. It uses the `projects_list` pipe from Tinybird, but that one does not touch activities, so we can't filter its data by activity type directly. The pipe gets its data from the `insightsProjects_filtered` pipe, which in turn gets its data from `insights_projects_populated_ds` data source, and that one is populated by a copy pipe, which does get activities data - but if we filter the data before it is copied, we affect every other pipe downstream from that one, which is no good.

---

> [!NOTE]
> Replaces isContribution filters with activity type-based filtering via new activityTypes_filtered pipe and updates related pipes/widgets to support includeCodeContributions/includeCollaborations.
> 
> - **Pipes: activity filtering overhaul**
>   - Replace `isContribution` checks with `AND (type, platform) IN (SELECT activityType, platform FROM `activityTypes_filtered`)` in:
>     - `activities_filtered.pipe`, `activities_filtered_historical_cutoff.pipe`, `activities_filtered_retention.pipe`
>     - `health_score_active_contributors.pipe`, `health_score_contributor_dependency.pipe`, `health_score_organization_dependency.pipe`
>     - `segmentId_aggregates_mv.pipe`
>   - Add params inherited from `activityTypes_filtered`: `includeCodeContributions`, `includeCollaborations` (and docs tweaks to `repos`).
> - **New pipe**
>   - `activityTypes_filtered.pipe`: parameterized filter over `activityTypes` with `includeCodeContributions` (default 1), `includeCollaborations`, `includeOtherContributions`; outputs `(activityType, platform)`.
> - **Widgets data sources**
>   - `top_member_org_copy.pipe`: constrain member/org activity counts to code contributions or collaborations using `activityTypes` (exclude other activity types).
> 
[...]

PR #3451, opened 2025-10-05
chore: add new pipes to filter data by activity types for the Top LF projects widget [IN-707]

borfast:
This is a follow-up to #3432, to fix the problem mentioned there, of not being able to use activity types to filter data for the "Top LF projects" widget.

It adds a new copy pipe that is similar to `insightsProjects_filtered` pipe but includes filtering for activity types, which we don't have on the former, and also uses some hard-coded filters for the Top LF projects widget to mimic the way the widget currently fetches data using dynamic filters.

These have to be hard-coded because this is a copy pipe and thus, we can't use dynamic filtering.

These pipes and corresponding data source exist just for the aforementioned widget, because adding the activity type filtering to insightsProjects_filtered could have too much of a performance hit, and that pipe is used in a lot of other places, so we would negatively impact the whole application.

### DIFF
--- a/services/libs/tinybird/datasources/projects_list_for_top_projects_widget_ds.datasource
+++ b/services/libs/tinybird/datasources/projects_list_for_top_projects_widget_ds.datasource
@@ -16,7 +16,6 @@ SCHEMA >
     `contributorCount` UInt64,
     `snapshotAt` DateTime64(3) DEFAULT now()
 
-
 ENGINE ReplacingMergeTree
 ENGINE_PARTITION_KEY tuple()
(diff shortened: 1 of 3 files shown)


borfast:
We decided to do this differently and filter on services/libs/tinybird/pipes/segmentId_aggregates_mv.pipe, which has the advantage that it covers other downstream pipes as well. The change was done [here](https://github.com/CrowdDotDev/crowd.dev/pull/3432/commits/10ed24a8b7f2dfbb36dd282664df24ca75050435).
Closing this one.

PR #3473, opened 2025-10-06
fix: filters applied in activity-timeline

joanagmaia:
### Changes proposed ✍️

- Place all filters inside the main `and` operator. Since we migrate to Tinybird we should make sure that the filters are applied always within an and or or operator.
- This will fix the applied filters for the people and organizations profile.

### DIFF
--- a/frontend/src/modules/activity/components/activity-timeline.vue
+++ b/frontend/src/modules/activity/components/activity-timeline.vue
@@ -311,19 +311,24 @@ const fetchActivities = async ({ reset } = { reset: false }) => {
   if (loading.value) {
     return;
   }
-  const filterToApply = {
-    platform: platform.value ? { in: [platform.value] } : undefined,
+
+  // Default filter to apply
+  const filterToApply: {
+    and: any[];
+  } = {
+    and: [
+      {
+        timestamp: {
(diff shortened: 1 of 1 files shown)


gaspergrom (approved):

ulemons (approved):

frontend/src/modules/activity/components/activity-timeline.vue
```diff
@@ -311,19 +311,24 @@ const fetchActivities = async ({ reset } = { reset: false }) => {
...
-  }
+  // Default filter to apply
+  const filterToApply: {
+    and: any[];
+  } = {
+    and: [
+      {
+        timestamp: {
+          gte: timestamp.value,
+        },
```
ulemons, line 323: @joanagmaia this only set the startDate, would it be reasonable to have a defaul smaller intervall to reduce the number of rows ?

joanagmaia: Hey, I think we could include it.
But perhaps let's wait for the Tinybird team to answer to this one first.
I'll merge this PR now to fix the filters in the profiles, and once they reply we can see if it makes sense to include a smaller range or not

PR #3509, opened 2025-10-14
fix: fix activityTypes Tinybird data source sorting key

borfast:
In the activityTypes Tinybird data source, we were missing the `activityType` field in the `ENGINE_SORTING_KEY`, which caused problems with deduplication.

This fixes it, and also removes isCodeContribution and isCollaboration from the sorting key to also prevent issues with duplicates, in case we change a setting in one of the activity types.

Note: there's no Jira ticket for this one, as this was an emergency fix yesterday.


gaspergrom (approved):
Please fix copilot comment

PR #3510, opened 2025-10-14
feat: add `lfx-internal-api` source to organization enrichment (CM-686)

skwowet:


services/apps/organizations_enrichment_worker/src/activities/enrichment.ts
```diff
@@ -1,421 +1,234 @@
...
-        value: (enriched as any).website,
-        verified: false,
-      })
-    }
-  }
+export async function getEnrichmentInput(
+  input: IEnrichableOrganization,
+): Promise<IOrganizationEnrichmentSourceInput> {
+  // dedup domains by value (keep first occurrence)
+  const uniqueDomains = uniqBy(input.identities, 'value')
```
themarolt, line 144: What about type? Usually we unique identities by value and type.

skwowet: Added type filter in the pgsql query used in [`fetchOrganizationEnrichment`](https://github.com/CrowdDotDev/crowd.dev/blob/feat/CM-686/services/libs/data-access-layer/src/organizations/enrichment/index.ts#L39-L65) to retrieve only verified primary domains from the db and deduplicate based on the value here.

themarolt (approved):

backend/src/database/migrations/R__organizationEnrichmentMaterializedViews.sql
```diff
@@ -0,0 +1,22 @@
+-- drop old materialized view if exists
+drop materialized view if exists "organizationsGlobalActivityCount" cascade;
+
+-- create the materialized view
+create materialized view "organizationsGlobalActivityCount" as
+select 
+    osa."organizationId",
+    sum(osa."activityCount") as total_count
```
epipav: NIT: maybe call this total_count_estimate since summing all these together won't give the real count bcs af overlaps

services/apps/members_enrichment_worker/src/workflows/enrichMember.ts
```diff
@@ -61,6 +62,9 @@ export async function enrichMember(
 
       const data = await getEnrichmentData(source, enrichmentInput)
 
+      // Record enrichment attempt
+      await touchMemberEnrichmentLastTriedAt(input.id)
```
epipav: The reason we were setting this one after processing is because this field is different than cache.updatedAt. 
cache.updatedAt always gets updated when we check (see bottom), but this one should only be updated when there's a change in the source data and the data is enrichable. So I think this should be in `processMemberSources`

services/apps/organizations_enrichment_worker/src/main.ts
```diff
@@ -30,7 +33,7 @@ export const svc = new ServiceWorker(config, options)
 setImmediate(async () => {
   await svc.init()
 
-  await scheduleOrganizationsEnrichment()
+  // await scheduleOrganizationsEnrichment()
```
epipav, line 36: NIT: let's not forget to uncomment this before merging

services/apps/organizations_enrichment_worker/src/workflows/enrichOrganization.ts
```diff
@@ -1,17 +1,92 @@
...
+  }
+
+  // Skipping credit check because this workflow uses a trusted internal API.
+  // If external sources are added in the future, credit checks should be implemented here.
+
+  // Fetch new enrichment data
+  const data = await getEnrichmentData(source, enrichmentInput)
+
+  // Record enrichment attempt
+  await touchOrganizationEnrichmentLastTriedAt(input.id)
```
epipav: same issue as before here, i think we should only set this after processing

services/libs/types/src/enums/organizations.ts
```diff
@@ -36,8 +36,9 @@ export enum OrganizationIdentityType {
 
 export enum OrganizationAttributeSource {
   CUSTOM = 'custom',
-  PDL = 'peopledatalabs',
-  EMAIL = 'email',
-  GITHUB = 'github',
+  ENRICHMENT_INTERNAL_API = 'enrichment-internal-api',
```
epipav: NIT: I think this naming is too vague - Internal API of what? There can be multiple internal APIs. If this is from LFX, we should add it to the naming or give it a more descriptive name

epipav (commented):
Overall lgtm, added a few comments

services/libs/types/src/enums/organizations.ts
skwowet: Yeah, you're right. This is from LFX. We can rename it to something more specific, like `ENRICHMENT_LFX_INTERNAL_API`, to make it clearer 👍🏻

services/apps/organizations_enrichment_worker/src/main.ts
skwowet: Yes

epipav (approved):

PR #3518, opened 2025-10-16
chore: handle gitlab rate limit errors better (CM-694)

themarolt:


services/libs/integrations/src/integrations/gitlab/api/getUser.ts
```diff
@@ -3,11 +3,16 @@ import { Gitlab, SimpleUserSchema, UserSchema } from '@gitbeaker/rest'
 import type { IProcessStreamContext } from '../../../types'
 import { RedisSemaphore } from '../utils/lock'
 
+import { handleGitlabError } from './errorHandler'
+
+const GITLAB_API_BASE_URL = 'https://gitlab.com/api/v4'
+export const GITLAB_GHOST_USER_ID = 1243277
```
mbani01, line 11: Is `GITLAB_GHOST_USER_ID` referring to a user not found? Can we have a comment here for better clarity?

mbani01 (approved):
Nicely done @themarolt 🙌 
LGTM, just left a minor comment

services/libs/integrations/src/integrations/gitlab/api/getUser.ts
themarolt: Added a comment - and this is for users that were deleted.

https://gitlab.com/ghost1 its a user id of this user.

```
About
This is a "Ghost User", created to hold all issues authored by users that have since been deleted. This user cannot be removed.
```

PR #3538, opened 2025-10-21
chore: remove is contribution from code [IN-708]

borfast:
Apologies in advance for the massive PR. I couldn't think of a nice way to break it down into multiple chunks. Fortunately, most of the files are changed in just deleting a line or two.

We want to remove the `isContribution` flag used in `activities` and `activityContributions` from everywhere.

This PR is the first step, where we just remove all the code that references the flag while still keeping the field in PostgreSQL and Tinybird.

With that in mind, we need to make sure any inserts to those tables will go through by providing a default value where necessary.

* Inserting `activities` implicitly uses the column's default value, and after the column is gone, nothing will be inserted
* Updating `activities` only sets keys that are explicitly given, so it's not a problem because we don't use the column anymore
* Inserting `activityRelations` uses a temporary value that should be removed afterwards
* Updating `activityRelations` only sets keys that are explicitly given, so it's not a problem

Using a temporary value is accomplished in [services/libs/data-access-layer/src/activities/sql.ts](https://github.com/CrowdDotDev/crowd.dev/pull/3538/files#diff-3a0ee54a4cd554b88e4cef9fb66e4edf1eeb829ab0b3645e93a0b3f7a3b88fa1) and [services/libs/data-access-layer/src/activities/insert.ts](https://github.com/CrowdDotDev/crowd.dev/pull/3538/files#diff-63becc327c3dbc4193a441f92fa2ed27938b53660c59272cf7f1cb9125937bce)

There are still quite a few references to that field in the code, namely:
* The migrations obviously still have it and will always have it unless we remove old migrations.
* Tinybird datasources, pipes and fixtures have been left untouched for now and will be updated when we actually remove the column from the databases.

[Jira ticket](https://linuxfoundation.atlassian.net/browse/IN-708)


package.json
```diff
@@ -7,5 +7,5 @@
     "@commitlint/config-conventional": "^19.8.0",
     "husky": "^9.1.7"
   },
-  "packageManager": "pnpm@9.11.0+sha512.0a203ffaed5a3f63242cd064c8fb5892366c103e328079318f78062f24ea8c9d50bc6a47aa3567cabefd824d170e78fa2745ed1f16b132e16436146b7688f19b"
+  "packageManager": "pnpm@10.18.3"
```
themarolt: Not sure about this one - we have a fixed pnpm version in our dockerfiles because there was a bug a while back with a newer version so we might just want to keep the old one as is...

borfast: Oh, crap, that was unintentional, it slipped through the cracks. I was having trouble installing dependencies and updated pnpm as an attempt to overcome the problem (didn't work, btw :sweat_smile:) but didn't mean for it to get here. Thanks for spotting it!

borfast: Fixed :heavy_check_mark:

PR #3555, opened 2025-10-29
fix: prevent foreign key violations during entity merges

skwowet:
## Changes

**1. Added override cleanup to role deletion**  
Modified `removeMemberRole` to delete associated overrides before removing the role, preventing foreign key violations.

**2. Refactored merge logic to preserve overrides**  
Restructured `mergeRoles` into a 3-phase approach:
- **Analyze**: Plan which roles to remove/add and track their original IDs
- **Remove**: Delete roles while storing their overrides for recreation  
- **Add**: Create new roles and intelligently transfer overrides to them (or to existing duplicate roles)

**3. Fixed role transformation**  
When moving a role from secondary to primary entity, it's now properly transformed with the target entity's ID and the original is removed, preventing orphaned records.

**4. Updated legacy code paths**  
Added override cleanup and transactions to older enrichment worker functions for consistency.

### DIFF
--- a/frontend/src/modules/member/pages/member-list-page.vue
+++ b/frontend/src/modules/member/pages/member-list-page.vue
@@ -142,20 +142,6 @@ const fetchMembersToMergeCount = () => {
 const loading = ref(true);
 const tableLoading = ref(true);
 
-const doGetMembersCount = () => {
-  (
-    MemberService.listMembers(
-      {
-        limit: 1,
-        offset: 0,
-      },
-      true,
-    ) as Promise<any>
-  ).then(({ count }) => {
-    membersCount.value = count;
-  });
-};
-
 const showLoading = (filter: any, body: any): boolean => {
   const saved: any = { ...savedFilterBody.value };
   delete saved.offset;
@@ -185,6 +171,10 @@ const fetch = ({
       limit: pagination.value.perPage,
       orderBy,
     },
+  }).then((result) => {
+    if (result && result.count !== undefined) {
+      membersCount.value = result.count;
+    }
   }).finally(() => {
     tableLoading.value = false;
     loading.value = false;
@@ -207,7 +197,6 @@ const onPaginationChange = ({
 
 onMounted(() => {
   fetchMembersToMergeCount();
-  doGetMembersCount();
   getMemberCustomAttributes();
   (window as any).analytics.page('Members');
 });


services/libs/data-access-layer/src/members/organizations.ts
```diff
@@ -424,6 +424,10 @@ export async function findNonIntersectingRoles(
 }
 
 export async function removeMemberRole(qx: QueryExecutor, role: IMemberOrganization) {
+  if (role.id) {
```
ulemons: Is it possible to not have this and only have the rest of the info ? `organizationId` and `memberId`? @skwowet

services/libs/data-access-layer/src/members/organizations.ts
```diff
@@ -700,65 +722,160 @@ export async function mergeRoles(
...
+
+        // If we found a primary override, use it, otherwise, use the first one
+        overrideToApply = primaryOverride?.override || relevantOverrides[0]?.override
+      }
+
+      if (overrideToApply) {
+        await changeOverride(qx, {
+          ...overrideToApply,
+          memberId: mergeStrat.targetMemberId(addRole),
+          memberOrganizationId: newRoleId,
```
ulemons: Is there a reason why we do this check outside the if at lin 763 ?

services/libs/data-access-layer/src/members/organizations.ts
```diff
@@ -700,65 +722,160 @@ export async function mergeRoles(
...
+          pr.title === addRole.title &&
+          ((pr.dateStart === null && addRole.dateStart === null) ||
+            (pr.dateStart &&
+              addRole.dateStart &&
+              new Date(pr.dateStart).getTime() === new Date(addRole.dateStart).getTime())) &&
+          ((pr.dateEnd === null && addRole.dateEnd === null) ||
+            (pr.dateEnd &&
+              addRole.dateEnd &&
+              new Date(pr.dateEnd).getTime() === new Date(addRole.dateEnd).getTime())),
+      )
```
ulemons: Do you think it can make sense to give meaningfull names to the different statements ?

skwowet: Yes

services/libs/data-access-layer/src/members/organizations.ts
skwowet: @ulemons do you mean a subquery?

```
DELETE FROM "memberOrganizationAffiliationOverrides"
WHERE "memberOrganizationId" IN (
  SELECT id FROM "memberOrganizations"
  WHERE "memberId" = $(memberId)
    AND "organizationId" = $(organizationId)
    AND "dateStart" = $(dateStart)
    AND "dateEnd" = $(dateEnd)
)
```

ulemons: Actually i was just wondering.. if role.id === null maybe we can have an early return here ? maybe not

services/libs/data-access-layer/src/members/organizations.ts
skwowet: good catch! not really a reason

I’ll push a fix for this 👍🏻

services/libs/data-access-layer/src/members/organizations.ts
skwowet: I added it as a subquery – I think that’s more consistent with our memberOrganization deletion.

themarolt (approved):

PR #3559, opened 2025-10-30
chore: add known bots to the list

joanagmaia:
- Adding very well known bots that are being missed to the list of known bots.


joanagmaia:
Canceling this PR as we already cover this in the strong pattern regex https://github.com/CrowdDotDev/crowd.dev/blob/main/services/libs/common_services/src/services/bot.service.ts#L7

PR #3560, opened 2025-10-30
chore: clean email providers list

skwowet:


skwowet:
@copilot check if the email providers list is accurate as per your knowledge and we accidentally don't include any legit company domains.

PR #3572, opened 2025-11-05
chore: github integration sync (CM-710)

themarolt:
- automatic github nango sync if organization adds repos we catch them and we add connection for that as wel
- re-enabled the github (v1) integration sync
- small fixes:
  - disable analytics tracking on local and if we don't have a `userId` as it was the case when a job does something that triggers analytics tracking (like `job-generator` job)
  - disable temporal encryption on local

### DIFF
--- a/backend/src/database/migrations/V1762773409__integrations-backup.sql
+++ b/backend/src/database/migrations/V1762773409__integrations-backup.sql
@@ -1,0 +1,10 @@
+create table "integrationsBackup" as
+select *
+from "integrations"
+where 1 = 2;
+
+alter table "integrationsBackup"
+    add column "backupCreatedAt" timestamptz not null default now();
+
+create index if not exists ix_integration_history_integration_id on "integrationsBackup" ("id");
+create index if not exists ix_integration_history_history_created_at on "integrationsBackup" ("backupCreatedAt");
--- a/services/apps/cron_service/src/jobs/integrationsBackup.job.ts
+++ b/services/apps/cron_service/src/jobs/integrationsBackup.job.ts
@@ -1,0 +1,24 @@
+import CronTime from 'cron-time-generator'
+
+import { IS_PROD_ENV } from '@crowd/common'
+import { WRITE_DB_CONFIG, getDbConnection } from '@crowd/data-access-layer/src/database'
+
+import { IJobDefinition } from '../types'
+
+const job: IJobDefinition = {
+  name: 'integrations-backup',
+  cronTime: CronTime.every(1).days(),
+  timeout: 60 * 60, // 1 hour = 60 * 60 seconds
+  enabled: async () => IS_PROD_ENV,
+  process: async (ctx) => {
+    ctx.log.info('Starting integrations backup job!')
+    const dbConnection = await getDbConnection(WRITE_DB_CONFIG(), 1, 0)
+    await dbConnection.query(`insert into "integrationsBackup" select * from "integrations";`)
+    await dbConnection.query(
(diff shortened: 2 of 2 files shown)


backend/src/api/integration/helpers/githubOrgRepos.ts
```diff
@@ -1,5 +1,6 @@
+import GithubIntegrationService from '@crowd/common_services/src/services/github.integration.service'
```
skwowet: nitpick: can we use an alias import instead of the full path?

backend/src/api/integration/helpers/githubSearchOrgs.ts
```diff
@@ -1,5 +1,6 @@
+import GithubIntegrationService from '@crowd/common_services/src/services/github.integration.service'
```
skwowet: same here

backend/src/api/integration/helpers/githubSearchRepos.ts
```diff
@@ -1,11 +1,12 @@
+import GithubIntegrationService from '@crowd/common_services/src/services/github.integration.service'
```
skwowet: same here

backend/src/services/collectionService.ts
```diff
@@ -1,6 +1,7 @@
 import { uniq } from 'lodash'
 
 import { getCleanString } from '@crowd/common'
+import GithubIntegrationService from '@crowd/common_services/src/services/github.integration.service'
```
skwowet: same here

backend/src/services/integrationService.ts
```diff
@@ -4,16 +4,17 @@ import { request } from '@octokit/request'
...
 
 import { EDITION, Error400, Error404, Error542 } from '@crowd/common'
+import { getGithubInstallationToken } from '@crowd/common_services'
 import {
   ICreateInsightsProject,
   deleteMissingSegmentRepositories,
   deleteSegmentRepositories,
   upsertSegmentRepositories,
 } from '@crowd/data-access-layer/src/collections'
+import { syncRepositoriesToGitV2 } from '@crowd/data-access-layer/src/integrations'
```
skwowet: same here

services/libs/data-access-layer/src/integrations/index.ts
```diff
@@ -437,6 +456,225 @@ export async function removeGitHubRepoMapping(
...
+      remotes: JSON.stringify(updatedRemotes),
+    },
+  )
+
+  log.info({ integrationId: gitIntegration.id, repoUrl }, 'Added repo to git integration settings!')
+
+  // Also sync to git.repositories table (git-integration V2)
+  await syncRepositoriesToGitV2(
+    qx,
+    [{ url: repoUrl, forkedFrom: null }],
```
mbani01: We need to determine whether the new repo is a fork to avoid reprocessing activities from the parent repo.
If we're getting repoUrl from nango, then we need to use GitHub API to get `forkedFrom`. We can re-use [getForkedFrom](https://github.com/CrowdDotDev/crowd.dev/blob/7f09a255c53745032c1d6031430ef3d73021e600/backend/src/services/githubIntegrationService.ts#L42C24-L42C37)

themarolt: @mbani01 I fixed it by passing forkedFrom from the caller function can you check if it's ok now?

mbani01: Thanks! LGTM

PR #3602, opened 2025-11-13
fix(integrations): seperate atlassian integrations connect and update (CM-738)

mbani01:
This pull request refactors the integration connection and update logic for both Confluence and Jira, making the API calls more explicit and improving maintainability. The main change is splitting the connect and update operations into separate service methods, and passing an optional `id` to distinguish between creating a new integration and updating an existing one. This affects both backend and frontend code, and also updates the relevant type definitions.

**Backend changes:**

* Separated connect and update logic for Confluence and Jira integrations in their respective helper files by calling either `connect*Integration` or `update*Integration` based on the presence of an `id` in the request body. [[1]](diffhunk://#diff-35ac7f354c3d73ead13d1d51b7f69533509a241a0419dafd770ea20759a931fcL7-R10) [[2]](diffhunk://#diff-3af2e5b2112a7669f0e03b51d43dae9cfa62d1da55ed4d062c9993da7904f92fL7-R10)
* Added optional `id` field to `ConfluenceIntegrationData` and `JiraIntegrationData` interfaces to support updates. [[1]](diffhunk://#diff-c54bf4fbf3c88afd88f3ca653efb859ab54a92e41eafc08307120822c270e791R2) [[2]](diffhunk://#diff-c1d5946b8a81463a4e65877ee37fe2cb88c5b83c6a3650b9c75ab6778568d32aR2)

**Frontend changes:**

* Updated the integration forms for Confluence and Jira to include the `id` field when calling the connect action, enabling update functionality. [[1]](diffhunk://#diff-82119bdaf5ef57292ba250aaffaf81beca22762937439ab818c1ae6cd5b10b96R247) [[2]](diffhunk://#diff-ce7c558d33289e3e0d31661439d4a7fe79c192c23df5dc9fd03f253fd534e962R253)
* Modified `IntegrationService` methods (`confluenceConnect` and `jiraConnect`) to accept an optional `id` parameter and include it in the API request payload. [[1]](diffhunk://#diff-e3824b8400bbe94af81b857c65934dedb083baa7e3534c4ad168e4d4842b2c5dL304-R306) [[2]](diffhunk://#diff-e3824b8400bbe94af81b857c65934dedb083baa7e3534c4ad168e4d4842b2c5dR432) [[3]](diffhunk://#diff-e3824b8400bbe94af81b857c65934dedb083baa7e3534c4ad168e4d4842b2c5dR441)
* Updated Vuex store actions (`doConfluenceConnect`, `doJiraConnect`) to pass the `id` parameter through to the service layer. [[1]](diffhunk://#diff-8b83462272f6fab878c7d2d63afa1207f20990660debf46dfd0f9971d5f3f9ddL582-R589) [[2]](diffhunk://#diff-8b83462272f6fab878c7d2d63afa1207f20990660debf46dfd0f9971d5f3f9ddR776) [[3]](diffhunk://#diff-8b83462272f6fab878c7d2d63afa1207f20990660debf46dfd0f9971d5f3f9ddR791)

### DIFF
--- a/services/libs/common_services/src/services/integration.service.ts
+++ b/services/libs/common_services/src/services/integration.service.ts
@@ -11,7 +11,7 @@ export class CommonIntegrationService {
   private static safeDecrypt(encryptedValue: string): string {
     try {
       return decryptData(encryptedValue)
-    } catch (error: any) {
+    } catch (error) {
       CommonIntegrationService.log.warn(`Failed to decrypt value: ${error?.message || error}`)
       return encryptedValue
     }
@@ -23,7 +23,7 @@ export class CommonIntegrationService {
    * @param settings - The settings object that may contain encrypted fields
    * @returns Settings object with decrypted values
    */
-  public static decryptIntegrationSettings(platform: string, settings: any): any {
+  public static decryptIntegrationSettings(platform: string, settings) {
     if (!settings) return settings
 
     switch (platform) {


backend/src/services/integrationService.ts
```diff
@@ -554,15 +554,67 @@ export default class IntegrationService {
     return IntegrationRepository.findGlobalIntegrationsStatusCount(args, this.options)
   }
 
+  private decryptIntegrationSettings(platform: string, settings: any): any {
```
themarolt: Can you perhaps just refactor this to a file in `common_services` lib? Eventually I want to move all the `integrationService.ts` there because there are lots of cases where we need to call these functions from `services` libs and apps and it's just easier if we start doing this with new things. You can just call it `integration.service.ts` and put it in `common_services` lib and just move the functions you added (related to encrypt/decrypt etc.) there not everything of course. Whatever is easy to move and doesn't take too much time :). The encrypt/decrypt from utils can also be moved to `common` library.

mbani01: Done ✅

PR #3619, opened 2025-11-20
feat: add be caching for api (CM-790)

ulemons:
# Add Backend Caching to Members API with Stale-While-Revalidate Strategy

## Overview

This PR implements a **Stale-While-Revalidate (SWR)** caching strategy for the members API to improve performance and user experience.

### 🔧 Refactored Query Architecture
- **Extracted `executeQuery` function**: Separated core query logic from caching logic for better maintainability
- **Added `IQueryMembersAdvancedParams` interface**: Type-safe parameter handling across all functions

### 🚀 Implemented SWR Caching Strategy
- **`MemberQueryCache`**: Redis-based caching with configurable TTL (6 hours)
- **Stale-While-Revalidate**: Returns cached data immediately while refreshing in background
- **Background refresh**: Non-blocking cache updates to ensure data freshness

### ⚡ Performance Benefits
- **Instant response times**: Always serves from cache when available
- **Improved UX**: No loading delays for users
- **Fresh data**: Automatic background updates keep cache current
- **Resilient**: Works even during database slowdowns

### Cache Strategy Flow
1. Check cache for existing data
2. If found: Return immediately + trigger background refresh
3. If not found: Execute query + cache result

### Background Refresh
```typescript
// Fire-and-forget background updates
void refreshCacheInBackground(qx, redis, cacheKey, params)

### DIFF
--- a/services/libs/data-access-layer/src/members/base.ts
+++ b/services/libs/data-access-layer/src/members/base.ts
@@ -199,7 +199,7 @@ export async function queryMembersAdvanced(
       attributeSettings,
     })
 
-    log.info(`Members advanced query cache hit: ${cacheKey}`)
+    log.debug(`Members advanced query cache hit: ${cacheKey}`)
     return cachedResult
   }
 
@@ -212,7 +212,7 @@ export async function queryMembersAdvanced(
       attributeSettings,
     })
 
-    log.info(`Members advanced count query cache hit: ${cacheKey}`)
+    log.debug(`Members advanced count query cache hit: ${cacheKey}`)
     return {
       rows: [],
       count: cachedCount,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: services/libs/data-access-layer/src/members/base.ts [hunk 1]
```diff
@@ -144,6 +165,110 @@ export async function queryMembersAdvanced(
...
+    log.info(`Members advanced count query cache hit: ${cacheKey}`)
+    return {
+      rows: [],
+      count: cachedCount,
+      limit,
+      offset,
+    }
+  }
+
+  log.info(`Executing members advanced query: ${cacheKey}`)
```
themarolt: Maybe just no info level logs on prod.

themarolt (approved):

PR #3625, opened 2025-11-20
fix: tinybird api rate limit (CM-749)

mbani01:
This pull request enhances the reliability of the `TinybirdClient` by introducing automatic retry logic for handling rate-limited (HTTP 429) responses from the Tinybird API. The changes ensure that temporary rate limits do not cause immediate failures, and instead, the client will wait and retry requests as recommended by the server. Additionally, logging is added for better observability. Below are the most important changes grouped by theme:

**Resilience and Retry Logic:**

* Added a `withRetry` method to `TinybirdClient` that wraps API requests and automatically retries them up to three times if a 429 rate limit is encountered, respecting the `Retry-After` header from the server.
* Updated all Tinybird API calls (`getPipe`, `query`, and `delete`) to use the new `withRetry` method, ensuring consistent retry behavior across all requests. [[1]](diffhunk://#diff-7f9677edd75db175346afc1e79530ace82dfd289b4b68015c83c654b2cb20b35L89-R140) [[2]](diffhunk://#diff-7f9677edd75db175346afc1e79530ace82dfd289b4b68015c83c654b2cb20b35L141-R195) [[3]](diffhunk://#diff-7f9677edd75db175346afc1e79530ace82dfd289b4b68015c83c654b2cb20b35L183-R239)
* Introduced a `sleep` helper method to pause execution between retries based on the server's suggested wait time.

**Logging and Observability:**

* Integrated a service-specific logger (`log`) to record warnings when rate limits are hit and retries are attempted, improving monitoring and debugging capabilities. [[1]](diffhunk://#diff-7f9677edd75db175346afc1e79530ace82dfd289b4b68015c83c654b2cb20b35R4-R7) [[2]](diffhunk://#diff-7f9677edd75db175346afc1e79530ace82dfd289b4b68015c83c654b2cb20b35R73-R113)

**Configuration:**

* Added a `maxRetries` property to the `TinybirdClient` class, defaulting to 3, to control the maximum number of retry attempts for rate-limited requests.

### DIFF
--- a/services/apps/script_executor_worker/src/bin/cleanup-fork-activities-and-maintainers.ts
+++ b/services/apps/script_executor_worker/src/bin/cleanup-fork-activities-and-maintainers.ts
@@ -61,8 +61,13 @@ interface DeletionStatus {
   error?: string
 }
 
-interface BatchResult {
-  batchNumber: string
+interface RepoCleanupResult {
+  repoUrl: string
+  status: 'success' | 'failure'
+  startTime: string
+  endTime: string
+  totalBatches: number
+  failedBatches: number
   deletions: {
     postgres: DeletionStatus
     tinybird: {
@@ -499,6 +504,10 @@ async function cleanupForkRepository(
   dryRun = false,
   tbToken?: string,
 ): Promise<void> {
+  const startTime = new Date().toISOString()
+  let failedBatches = 0
+  let totalBatches = 0
+
   if (dryRun) {
     log.info(`\n${'='.repeat(80)}`)
     log.info(`[DRY RUN MODE] Processing: ${repo.url}`)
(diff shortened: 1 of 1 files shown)


services/libs/database/src/tinybirdClient.ts
```diff
@@ -138,11 +186,13 @@ export class TinybirdClient {
     const url = `${this.host}/v0/sql`
     const body: Record<string, unknown> = { q: query, ...bodyParams }
 
-    const result = await axios.post<T>(url, body, {
-      headers: this.getHeaders('application/json'),
-      responseType: 'json',
-      httpsAgent: TinybirdClient.httpsAgent,
-    })
+    const result = await this.withRetry(async () =>
```
ulemons: In both cases (`this.pipe` and `this.pipeSql`) we are using the `withRetry` function, so we are not giving the possibility to not use the retry. 
Do you think we can have a parameter here saying like withRetry = true by default? if i set withRetry false we could simply have the `maxRetries = 1`.
what do you think ?

services/libs/database/src/tinybirdClient.ts
```diff
@@ -65,6 +70,47 @@ export class TinybirdClient {
...
+  private async withRetry<T>(fn: () => Promise<T>): Promise<T> {
+    let lastError: Error | undefined
+
+    for (let attempt = 0; attempt <= this.maxRetries; attempt++) {
+      try {
+        return await fn()
+      } catch (error: any) {
+        lastError = error
+
+        if (error.response?.status === 429 && attempt < this.maxRetries) {
```
ulemons: i wanted to implment the same but for the status 408, do you think we can have a list of statuses here ?

services/libs/database/src/tinybirdClient.ts
mbani01: Yes makes sense

services/libs/database/src/tinybirdClient.ts
mbani01: Sure 👌

PR #3626, opened 2025-11-20
feat: adding index for organizations (CM-791)

ulemons:
## Changes:
add same index was added on the member query also on the organization query


backend/src/database/repositories/organizationRepository.ts
```diff
@@ -1687,6 +1690,9 @@ class OrganizationRepository {
 
     const countQuery = createQuery('COUNT(*)')
 
+    log.info('Executing organization find query: ', { query, params })
```
themarolt: Maybe just no info logs when deployed to prod.

PR #3647, opened 2025-11-26
chore: update GitHub nango onboarding interval (CM-793)

mbani01:
This pull request updates the GitHub integration workflow in the Nango worker service to improve connection handling and onboarding pacing. The main changes include **switching to a fixed onboarding interval of 6 minutes** and extending activity timeouts (just to be safe).

**Improvements to connection management:**

* Parallelized deletion operations for unused and duplicate connections by grouping related activity calls with `Promise.all`, improving efficiency and reducing overall deletion time.

**Workflow and onboarding adjustments:**

* Increased the `startToCloseTimeout` for activities from 1 hour to 2 hours, providing more time for longer-running tasks.
* Replaced the random onboarding delay (1–5 minutes) with a fixed interval of 6 minutes between repository onboardings to ensure a more predictable and even distribution throughout the day.
* Introduced the `REPO_ONBOARDING_INTERVAL_MINUTES` constant to manage the fixed onboarding interval.

### DIFF
--- a/services/apps/nango_worker/src/workflows/syncGithubIntegration.ts
+++ b/services/apps/nango_worker/src/workflows/syncGithubIntegration.ts
@@ -18,25 +18,25 @@ export async function syncGithubIntegration(args: ISyncGithubIntegrationArgument
 
     // delete connections that are no longer needed
     for (const repo of result.reposToDelete) {
-      await Promise.all([
-        // delete nango connection
-        activity.deleteConnection(integrationId, result.providerConfigKey, repo.connectionId),
-        // delete connection from integrations.settings.nangoMapping object
-        activity.removeGithubConnection(integrationId, repo.connectionId),
-        // delete githubRepos mapping
-        activity.unmapGithubRepo(integrationId, repo.repo),
-      ])
+      // delete nango connection
+      await activity.deleteConnection(integrationId, result.providerConfigKey, repo.connectionId)
+
+      // delete connection from integrations.settings.nangoMapping object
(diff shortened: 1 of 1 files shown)


services/apps/nango_worker/src/workflows/syncGithubIntegration.ts
```diff
@@ -16,25 +18,25 @@ export async function syncGithubIntegration(args: ISyncGithubIntegrationArgument
...
-      await activity.removeGithubConnection(integrationId, repo.connectionId)
-
-      // we don't unmap because this one was duplicated
+      await Promise.all([
+        // delete nango connection
+        activity.deleteConnection(integrationId, result.providerConfigKey, repo.connectionId),
+        // delete connection from integrations.settings.nangoMapping object
+        activity.removeGithubConnection(integrationId, repo.connectionId),
+        // we don't unmap because this one was duplicated
+      ])
```
mbani01: @themarolt can you confirm if it's safe to trigger the clenaup operations concurrently? I see they are not related, and thought we can speed it up to reduce extra delays WDYT?

themarolt: If delete connection from nango fails we shouldn't delete the `nangoMapping` part. So we can keep track what is on nango and what is not.

So we can concurrently process `duplicatesTODelete` but not individual operations - `activity.deleteConnection` should always happen before `activity.removeGithubConnection`.

mbani01: You're right! I disabled it

PR #3673, opened 2025-12-04
fix: nango repos to delete

joanagmaia:
**Issue**
We disconnected github-nango from [OpenStack](https://cm.lfx.dev/integrations/1dd150b2-c066-4b61-b36a-b5dc9b7dc0b0/ce62cbde-b5d3-4f94-a085-0938e5980510), there were close to 700 repos for this organization.

In the code where we dele[te nango connections automatically when they are no longer being used ](https://github.com/CrowdDotDev/crowd.dev/blob/f486e7d9a41f72ca831e91211bc5918ae99e57c5/services/apps/nango_worker/src/activities/nangoActivities.ts#L223), we actually only look for integrations that are not softly deleted. So the issue is that we are actually only covering the cases where we disconnect one single repo from an integration that is still connected. We are not deleting nango connection for entire integrations that were disconnected. This can save us a lot of connections, that are no longer needed and are simply consuming resources.

**Fix**
- Add a method to fetch deleted integrations by id.
- If there is a match, add all the repos from that integration to `reposToDelete`

---

> [!NOTE]
> Include deleted GitHub-Nango integrations when computing connections to delete, adding a DAL query to fetch soft-deleted integrations.
> 
> - **Nango worker (`services/apps/nango_worker/src/activities/nangoActivities.ts`)**
>   - Enhance `analyzeGithubIntegration` to handle missing integrations by fetching soft-deleted ones via `fetchDeletedIntegrationById`.
>   - When a deleted integration is GitHub Nango, aggregate all `settings.nangoMapping` entries into `reposToDelete` and log the count.
> - **Data Access Layer (`services/libs/data-access-layer/src/integrations/index.ts`)**
>   - Add `fetchDeletedIntegrationById` to query integrations where `deletedAt` is not null (returns `id`, `platform`, `settings`, `segmentId`).
>   - Export and consume the new function from the worker.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37456397fb9b8eb15c5c2ddaa0d77240e30ad06e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2199,8 +2199,8 @@ importers:
         specifier: workspace:*
         version: link:../logging
       '@slack/webhook':
-        specifier: ^7.0.2
-        version: 7.0.6
+        specifier: ^6.1.0
+        version: 6.1.0
     devDependencies:
       '@types/node':
         specifier: ^18.16.3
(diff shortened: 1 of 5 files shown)


services/apps/nango_worker/src/activities/nangoActivities.ts
```diff
@@ -221,6 +222,10 @@ export async function analyzeGithubIntegration(
   const reposToSync: IGithubRepoData[] = []
 
   const integration = await fetchIntegrationById(dbStoreQx(svc.postgres.writer), integrationId)
+  const deletedIntegration = await fetchDeletedIntegrationById(
```
themarolt: Load this in an else only after you know that `integration` is null/undefined.

themarolt (changes requested):

PR #3682, opened 2025-12-09
fix: implement and trigger member cache invalidation after update ops [CM-823]

mbani01:
This pull request introduces a mechanism to automatically invalidate the member query cache in key member-related operations, improving data consistency across the application. The changes primarily add the `MemberQueryCache.invalidateAll()` method and ensure it is called after unmerge, update, and bulk delete operations in `MemberService`.

### Member query cache invalidation

* Added the `invalidateAll()` method to the `MemberQueryCache` class, which deletes all cached member query results and counts, logging the outcome and handling errors gracefully.
* Exported the `MemberQueryCache` class from the data-access-layer to make it available for use in other modules.
* Imported `MemberQueryCache` into `memberService.ts` to enable cache invalidation functionality.

### MemberService integration

* Implemented a private `invalidateMemberQueryCache` helper in `MemberService` to invoke cache invalidation, with error logging that does not interrupt the main operation flow.
* Called `invalidateMemberQueryCache` after member unmerge, update, and bulk delete operations to ensure stale cache entries are removed whenever member data is changed. [[1]](diffhunk://#diff-b3bd810e62d7a283a7a0e50043fe8b8297e449538c2a1ff0077b00c69e0e1f3eR760-R762) [[2]](diffhunk://#diff-b3bd810e62d7a283a7a0e50043fe8b8297e449538c2a1ff0077b00c69e0e1f3eR1271-R1273) [[3]](diffhunk://#diff-b3bd810e62d7a283a7a0e50043fe8b8297e449538c2a1ff0077b00c69e0e1f3eR1325-R1327)

### DIFF
--- a/backend/src/api/member/memberMerge.ts
+++ b/backend/src/api/member/memberMerge.ts
@@ -1,5 +1,5 @@
 import { CommonMemberService } from '@crowd/common_services'
-import { optionsQx } from '@crowd/data-access-layer'
+import { MemberQueryCache, optionsQx } from '@crowd/data-access-layer'
 
 import Permissions from '../../security/permissions'
 import track from '../../segment/track'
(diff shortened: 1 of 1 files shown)


backend/src/services/memberService.ts
```diff
@@ -67,6 +68,17 @@ import SettingsService from './settingsService'
 export default class MemberService extends LoggerBase {
   options: IServiceOptions
 
+  private async invalidateMemberQueryCache(): Promise<void> {
+    try {
+      const cache = new MemberQueryCache(this.options.redis)
+      await cache.invalidateAll()
```
ulemons, line 74: @mbani01 did you tested this invalidate only the "members-advanced" cache ?

ulemons:
@mbani01 this make sense because of the bug, maybe we can think of having also the memberId/orgId in the prefix of the key so we could avoid to remove everything everytime we update something

backend/src/services/memberService.ts
mbani01: No, but I think it will cause some inconsistencies, especially with merge/unmerge and delete operations

PR #3722, opened 2025-12-30
feat(std-repositories-tables): services write to new tables  [CM-865]

mbani01:
This pull request introduces significant improvements to the repository integration and synchronization logic within the `IntegrationService` in `backend/src/services/integrationService.ts`. The main focus is on ensuring that repository mappings between source integrations and the unified `public.repositories` table are consistent, robust, and transactional across different platforms (GitHub, GitLab, Gerrit, and direct Git). The changes add validation, transactional safety, and better handling of edge cases when mapping, restoring, or soft-deleting repositories.

Key changes include:

**Repository Mapping and Synchronization Enhancements**
- Added a new method `mapUnifiedRepositories`, which handles inserting, restoring, and soft-deleting repositories in the unified `public.repositories` table in a transactional manner. This includes robust validation to prevent remapping repositories across different integrations and ensures consistency across platforms.
- Introduced helper methods like `validateRepoIntegrationMapping` and `buildRepositoryPayloads` to support the main mapping logic, including building payloads with correct associations (segment, integration, insights project, forkedFrom, etc.).

**Platform-Aware Repository Synchronization**
- Updated the `gitConnectOrUpdate` method and all relevant integration flows (GitHub, GitLab, Gerrit, direct Git) to call `mapUnifiedRepositories` with the correct platform and integration context, ensuring that repository synchronization is handled appropriately for each platform. [[1]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR961-R963) [[2]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdL1403-R1439) [[3]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdL1731-R1790)
- Added a `sourcePlatform` parameter to `gitConnectOrUpdate` to control when unified repository mapping should be triggered, allowing for more flexible integration flows.

**Integration with Data Access Layer**
- Expanded imports from the data access layer to support new repository operations, such as fetching, inserting, restoring, and soft-deleting repositories.

**Improved Transaction Management**
- Ensured that all repository mapping operations are performed within a transaction, with proper commit and rollback handling to maintain data integrity in case of errors.

**Platform-Specific Handling in Integration Flows**
- Updated integration flows for GitHub, GitLab, Gerrit, and direct Git to pass the appropriate platform type and ensure that repository synchronization logic is platform-aware. [[1]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR275) [[2]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR465) [[3]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR1072) [[4]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR1084) [[5]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR2866) [[6]](diffhunk://#diff-80e1c7e63674653fe2cf5bcbe4a2212499d72014f328ad2e6b617146f0a256fdR2877)

These changes make the repository integration process more reliable, maintainable, and scalable across different source platforms.

---

> [!NOTE]
> Consolidates repository mapping for all code platforms into `public.repositories` with transactional insert/restore/soft-delete and platform-aware validation.
> 
> - Adds `GET /integration/:id/repositories` and removes legacy per-platform GETs; frontend switches to `IntegrationService.fetchIntegrationRepositories` for GitHub/GitLab/Git
> - Implements `IntegrationService.mapUnifiedRepositories` and helpers; wires calls from GitHub, GitLab, Gerrit, and direct Git flows; enforces ownership/mirroring rules and improves deletion logic (including Git integration safeguards)
> - Enhances `gitConnectOrUpdate` to coordinate with unified mapping and prune owned `git.repositories`; updates `syncRepositoriesToGitV2` to reuse IDs (incl. soft-deleted)
> - Introduces `data-access-layer/src/repositories` module (insert/restore/soft-delete/upsert/query/mapping)
> - Nango worker updates: upsert/soft-delete repos in `public.repositories` during GitHub sync; cronjob now also updates `public.repositories`
> - UI: Git/Gerrit settings reflect unified mappings; Git disables edits for mirrored repos; Gerrit defaults `enableGit` true and minor UX tweaks
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ca3693e96d64f125b03bc6c7dce4b5ae7b14d941. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/backend/src/services/integrationService.ts
+++ b/backend/src/services/integrationService.ts
@@ -1827,7 +1827,7 @@ export default class IntegrationService {
 
       // Build full repository URLs from orgURL and repo names
       const currentSegmentId = this.options.currentSegments[0].id
-      const remotes = integrationData.remote.repoNames.map((repoName) => {
+      let remotes = integrationData.remote.repoNames.map((repoName) => {
         const fullUrl = stripGit(`${integrationData.remote.orgURL}/${repoName}`)
         return { url: fullUrl, forkedFrom: null }
       })
@@ -1871,6 +1871,12 @@ export default class IntegrationService {
       if (integrationData.remote.enableAllRepos) {
         integrationData.remote.repoNames = res
       }
+
+      // Rebuild remotes after enableAllRepos may have updated repoNames
+      remotes = integrationData.remote.repoNames.map((repoName) => {
+        const fullUrl = stripGit(`${integrationData.remote.orgURL}/${repoName}`)
+        return { url: fullUrl, forkedFrom: null }
+      })
 
       connectionId = await createNangoConnection(NangoIntegration.GERRIT, {
         params: {


services/libs/data-access-layer/src/repositories/index.ts
```diff
@@ -0,0 +1,374 @@
...
+      archived boolean,
+      "forkedFrom" text,
+      excluded boolean
+    )
+    WHERE r.url = v.url
+    `,
+    { values: JSON.stringify(values) },
+  )
+
+  // Reset git.repositoryProcessing for git re-onboarding
```
ulemons, line 311: shouldnt we handle these two calls as a single transaction ? what if one fails and the other not?

services/libs/data-access-layer/src/integrations/index.ts
```diff
@@ -550,62 +550,63 @@ export async function syncRepositoriesToGitV2(
...
+  // Include soft-deleted repos to reuse their IDs on reconnect
   const existingRepos: Array<{
     id: string
     url: string
   }> = await qx.select(
     `
     WITH github_repos AS (
       SELECT id, url FROM "githubRepos" 
-      WHERE url IN ($(urls:csv)) AND "deletedAt" IS NULL
+      WHERE url IN ($(urls:csv))
```
ulemons, line 564: why are we removing the deletedAt filter ?

services/libs/data-access-layer/src/repositories/index.ts
```diff
@@ -0,0 +1,374 @@
...
+  forkedFrom?: string | null
+  excluded?: boolean
+}
+
+/**
+ * Bulk inserts repositories into public.repositories and git.repositoryProcessing
+ * @param qx - Query executor (should be transactional)
+ * @param repositories - Array of repositories to insert
+ */
+export async function insertRepositories(
```
ulemons, line 39: I see we don't have any `on conflict` clause, what if we try to reinsert again ? is it something allowed ?

services/libs/data-access-layer/src/integrations/index.ts
mbani01: There was a bug, when we reconnect a deleted repo, it fails because we try to create a new one with new ID.
While we should keep same ID as it's being referenced on MaintainersInternal table.
That's why removing the deletedAt is needed to keep same id on git.repositories instead of trying to insert a new one.
Anyway, this logic will be deprecated soon post migration and we'll only support `public.repositories` table which I believe is handling this logic simpler (using [restoreRepositories](https://github.com/CrowdDotDev/crowd.dev/pull/3722/changes/BASE..74a31d91bd664495783e04822d3d4f431d3b754a#diff-782eb8bead5490c87894035daf29503178742d5e713170fcaa416f78b95ce17dR259))

There is a comment above to avoid confusion:  `// Include soft-deleted repos to reuse their IDs on reconnect`

services/libs/data-access-layer/src/repositories/index.ts
mbani01: Blindly re-inserting repos shouldn’t be allowed, as it can have side effects. I tried to decouple insert and restore (re-inserting deleted repos) operations, since their logic isn’t exactly the same.

This shouldn’t be problematic for now, as all detection and application of insert/restore operations are handled in mapUnifiedRepositories (which isn’t ideal, but I’ll revisit it after the migration)

services/libs/data-access-layer/src/repositories/index.ts
mbani01: The caller (mapUnifiedRepositories) already manages the transaction

PR #3902, opened 2026-03-10
feat: added project id from the results returned by the search endpoint

emlimlf:
## In this PR

Added the project id from the results returned by the search endpoint. This will be used by the create collection since it requires the ID instead of the slug

## Ticket
[IN-1010](https://linuxfoundation.atlassian.net/browse/IN-1010)

[IN-1010]: https://linuxfoundation.atlassian.net/browse/IN-1010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Adds a nullable/empty `id` column to an existing search pipe response with no auth or data-mutation logic; consumers must tolerate the new field.
> 
> **Overview**
> The unified Tinybird search pipe now returns an **`id`** field on every row so clients can use the real **project ID** (not only slug) when wiring flows like **create collection**.
> 
> **Project** hits expose `insightsProjects_filtered.id`; **collection** and **repository** rows use an empty `id` so the `UNION ALL` column shape stays aligned. Pipe docs were updated to describe the new field.
> 
> There is also a **format-only** change to the `--export-name` validation block in `triggerBootstrap.ts` (no behavior change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e40e9eec145aea2355f954618ba7f261977e5e29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joanagmaia:
@emlimlf @gaspergrom is this something that's still needed? It wasn't reviewed and it seems that it's stale. If this is already in production, let's make sure we merge this

gaspergrom (approved):

gaspergrom (dismissed):

PR #4006, opened 2026-04-07
feat: pcc sync worker (CM-1086)

themarolt:
> [!NOTE]
> **Medium Risk**
> Adds a new worker that performs production DB updates to `segments`/`insightsProjects` and introduces new job-claiming/filtering semantics in `MetadataStore`, which could affect Snowflake job processing and cleanup behavior if misconfigured.
> 
> **Overview**
> Introduces a new `pcc_sync_worker` service that schedules daily Snowflake exports/cleanup via Temporal, streams exported Parquet files from S3, and **syncs PCC project metadata into CDP** by updating matching `segments` and creating/updating `insightsProjects`.
> 
> Adds DB support for the sync by introducing a nullable `segments.maturity` column and a new `pcc_projects_sync_errors` table with deduping indexes for tracking schema issues, hierarchy/slug mismatches, and name conflicts.
> 
> Refactors the shared `@crowd/snowflake` library to centralize `MetadataStore`, `S3Service`, and `SnowflakeExporter` (including `buildS3FilenamePrefix`), and updates `snowflake_connectors` to use these exports plus **platform-filtered** job claiming/cleanup (with a new `releaseClaim` flow and optional skipped-count cleanup gating).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15bb77c4cf947f6698f2ee1d7d411ef967560ce3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/apps/pcc_sync_worker/src/activities/exportActivity.ts
```diff
@@ -0,0 +1,100 @@
...
+    log.info({ platform: PLATFORM, sourceName: SOURCE_NAME }, 'PCC export completed')
+  } catch (err) {
+    log.error({ platform: PLATFORM, sourceName: SOURCE_NAME, err }, 'PCC export failed')
+    throw err
+  } finally {
+    await exporter
+      .destroy()
+      .catch((err) => log.warn({ err }, 'Failed to close Snowflake connection'))
+  }
+}
```
mbani01: This seems duplicated across snowflake_connectors/PCC_sync, can we move these to @crowd/snowflake as well?

services/libs/snowflake/src/metadataStore.ts
```diff
@@ -54,15 +54,27 @@ export class MetadataStore {
...
+  ): Promise<SnowflakeExportJob | null> {
+    let platformFilter = ''
+    let params: Record<string, unknown> = {}
+    if (platform) {
+      platformFilter = 'AND platform = $(platform)'
+      params = { platform }
+    } else if (platforms && platforms.length > 0) {
+      platformFilter = 'AND platform = ANY($(platforms)::text[])'
+      params = { platforms }
+    }
```
mbani01: The platform / platforms arguments are easy to misuse.
Instead of passing platform(s) directly, we can build the clause and params and pass them as argument:

```suggestion
interface PlatformFilter {
  clause: string
  params: Record<string, unknown>
}
function buildPlatformFilter(platforms: string[]): PlatformFilter {
  if (platforms.length === 0) {
    return { clause: 'AND FALSE', params: {} }
  }
  return {
    clause: 'AND platform = ANY($(platforms)::text[])',
    params: { platforms },
  }
}
  async claimOldestPendingJob(filter: PlatformFilter): Promise<SnowflakeExportJob | null> {
```
WDYT ?

mbani01 (commented):
Well done @themarolt 💪  left a couple comments

services/apps/pcc_sync_worker/src/activities/cleanupActivity.ts
```diff
@@ -0,0 +1,33 @@
+import { WRITE_DB_CONFIG, getDbConnection } from '@crowd/database'
+import { getServiceChildLogger } from '@crowd/logging'
+import { SlackChannel, SlackPersona, sendSlackNotification } from '@crowd/slack'
+import { MetadataStore, S3Service } from '@crowd/snowflake'
+
+const log = getServiceChildLogger('cleanupActivity')
+
+const PLATFORM = 'pcc'
+
+export async function executeCleanup(intervalHours = 24): Promise<void> {
```
joanagmaia, line 10: Eric suggested this https://docs.google.com/document/d/1t6HyZdHGM9TA47fyJ5jRQ96z4O5esmnCDu03yNS4lwI/edit?disco=AAAB2hXY2p8, to basically explore the S3 bucket policies for automatically purging old snapshots / query-results. Perhaps leveraging that we could avoid having our own internal job to clean up S3 buckets.
Let me know if it's something worth checking or if it would still be preferable to have this cleanup job.

services/apps/pcc_sync_worker/src/activities/exportActivity.ts
```diff
@@ -0,0 +1,100 @@
...
+import { WRITE_DB_CONFIG, getDbConnection } from '@crowd/database'
+import { getServiceChildLogger } from '@crowd/logging'
+import { MetadataStore, SnowflakeExporter } from '@crowd/snowflake'
+
+const log = getServiceChildLogger('exportActivity')
+
+const PLATFORM = 'pcc'
+const SOURCE_NAME = 'project-hierarchy'
+
+function buildSourceQuery(): string {
```
joanagmaia, line 19: Hey, I know this is the query I used before on the first data analysis, but this is not following the latest mapping rules of the new PROJECTS_SPINE table which already has the hierarchy defined -> We don't need to do it ourselves.

Proposed new query:
```
SELECT
    p.project_id,
    p.name,
    p.description,
    p.project_logo,
    p.project_status,
    p.project_maturity_level,
    ps.mapped_project_id,
    ps.mapped_project_name,
    ps.mapped_project_slug,
    ps.hierarchy_level,
    s.segment_id
FROM ANALYTICS.SILVER_DIM.PROJECTS p
LEFT JOIN ANALYTICS.SILVER_DIM.PROJECT_SPINE ps ON ps.base_project_id = p.project_id
LEFT JOIN ANALYTICS.SILVER_DIM.ACTIVE_SEGMENTS s
    ON s.source_id = p.project_id
    AND s.project_type = 'subproject'
WHERE p.project_id NOT IN (
    SELECT DISTINCT parent_id
    FROM ANALYTICS.SILVER_DIM.PROJECTS
    WHERE parent_id IS NOT NULL
)
ORDER BY p.name, ps.hierarchy_level ASC;
```

**Important**
Level/depth is now longer in a column but rather in new rows. E.g.
```

project_id | name | mapped_project_id | mapped_project_name | mapped_project_slug | hierarchy_level | segment_id
----------- | -------- | ------------------- | ----------------------- | ---------------------- | --------------- | 
kubectl_id  | kubectl  | kubectl_id | kubectl | kubectl | 1 | seg_123
kubectl_id  | kubectl  | kubernetes_id | kubernetes | kubernetes | 2 | seg_123
kubectl_id  | kubectl  | tlf_id | tlf | tlf | 3 | seg_123
```

Changes:
**1. No hardcoded depth limit**
The recursive CTE hard-codes depth_1 through depth_5. If the hierarchy ever grows to 6+ levels, we need a schema change. The new query handles any depth automatically via additional rows.

**2. Reuses existing model infrastructure**
PROJECT_SPINE already computes and materializes the hierarchy. The recursive CTE duplicates that logic inline — any bug fix or change to hierarchy traversal would need to be made in two places.

**3. Simpler, more readable SQL**
The recursive CTE is ~20 lines of stateful logic that requires understanding how the recursion builds depth_N columns. The new query is a straightforward set of joins that's immediately understandable.

**5. Normalized shape**
Wide columns (depth_1..5) are harder to work with downstream, filtering, aggregating, or displaying "what level is this?" requires knowing which column to look at. One row per level is easier to GROUP BY, FILTER, or JOIN against.

**6. Removed repository_url on the leaf project**
Since we're no longer using repository_url to automatically onboard, as that's a free text url in the UI and we shouldn't rely on it, then we don't need to return it as well.

services/apps/pcc_sync_worker/src/consumer/pccProjectConsumer.ts
```diff
@@ -0,0 +1,498 @@
...
+  //   eff=3 or 4 → only the subproject segment carries this PROJECT_ID
+  // So this UPDATE always writes the correct name and never touches unrelated levels.
+  await db.none(
+    `UPDATE segments
+     SET name        = $(name),
+         status      = $(status)::"segmentsStatus_type",
+         maturity    = $(maturity),
+         description = $(description),
+         "updatedAt" = NOW()
+     WHERE "sourceId" = $(sourceId) AND "tenantId" = $(tenantId)`,
```
joanagmaia, line 474: Missing logo?

services/apps/pcc_sync_worker/src/parser/rowParser.ts
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * Depth > 4 (raw > 5): SCHEMA_MISMATCH — surfaced to pcc_projects_sync_errors.
+ */
+import type { MappingRule, ParseResult, PccParquetRow } from './types'
+
+/**
+ * PCC PROJECT_STATUS → CDP segmentsStatus_type enum.
+ * CDP enum values: active | archived | formation | prospect
+ */
+const STATUS_MAP: Record<string, string> = {
```
joanagmaia: Should we use SegmentStatus enum for the mapped values? Just to only have it defined in one place?

services/apps/pcc_sync_worker/src/parser/rowParser.ts
```diff
@@ -0,0 +1,185 @@
...
+  'Formation - On Hold': 'formation',
+  Prospect: 'prospect',
+}
+
+/**
+ * Intermediate PCC nodes that are transparent in the CDP hierarchy.
+ * When D2 equals one of these, it is skipped and D1 ("The Linux Foundation")
+ * is used as the CDP project group instead.
+ */
+const TRANSPARENT_INTERMEDIATES = new Set(['LF Projects, LLC'])
```
joanagmaia: Until I get validation on this https://docs.google.com/document/d/1t6HyZdHGM9TA47fyJ5jRQ96z4O5esmnCDu03yNS4lwI/edit?tab=t.0#heading=h.i6ky9k93x8ma let's not add a rule for these. Let's remove it

services/apps/pcc_sync_worker/src/parser/rowParser.ts
```diff
@@ -0,0 +1,185 @@
...
+        name,
+      },
+    }
+  }
+
+  const d1 = row.DEPTH_1 ?? null
+  const d2 = row.DEPTH_2 ?? null
+  const d3 = row.DEPTH_3 ?? null
+  const d4 = row.DEPTH_4 ?? null
+  const d5 = row.DEPTH_5 ?? null
```
joanagmaia: Fix depth parsing based on this comment https://github.com/linuxfoundation/crowd.dev/pull/4006/changes#r3086073480

joanagmaia (changes requested):
Main thing to update is the query to snowflake which in return will affect how we detect hierarchy mismatches. Main changes are:
- Use PROJECTS_SPINE to rely on depth mapping
- Do not hard code depth level up to 5, simply flat depth into multiple rows

services/apps/pcc_sync_worker/src/activities/exportActivity.ts
themarolt: done

services/libs/snowflake/src/metadataStore.ts
themarolt: done

services/apps/pcc_sync_worker/src/activities/cleanupActivity.ts
themarolt: I would suggest doing this after this is merged and deployed as it's the same for the snowflake connector right now. We can change this for both of these when we need it but I would say the cleanup with the automatic purge on the AWS side is also a bit problematic if we don't process the data in time if the service is down the data gets deleted probably on a time since created basis. But then again if this happens we can just import the data again and process that batch. Still I would do this separate of this PR and for both services at once pcc-sync-worker and snowflake-connector.

services/apps/pcc_sync_worker/src/activities/exportActivity.ts
themarolt: Done just testing it now before resolving the comment.

services/apps/pcc_sync_worker/src/consumer/pccProjectConsumer.ts
themarolt: Logo is in `insightsProjects` table - we can also add it to `segments` table if you want via migration and the code change but it's gonna be the same data no?

services/apps/pcc_sync_worker/src/parser/rowParser.ts
themarolt: Done.

services/apps/pcc_sync_worker/src/parser/rowParser.ts
themarolt: Done.

services/apps/pcc_sync_worker/src/parser/rowParser.ts
themarolt: Removed — no intermediate-skipping logic in the new parser.

services/apps/pcc_sync_worker/src/activities/cleanupActivity.ts
joanagmaia: Sounds good. Created a task to address it in the future: https://linuxfoundation.atlassian.net/browse/CM-1127

services/apps/pcc_sync_worker/src/consumer/pccProjectConsumer.ts
joanagmaia: Not sure why but I thought we also add a `logo` column in segments table, tha's why I suggested to also update there. Nevermind, no need to update anything here

services/apps/pcc_sync_worker/src/consumer/pccProjectConsumer.ts
```diff
@@ -0,0 +1,584 @@
...
+
+  // No conflict — update all active insightsProject rows linked to any segment that
+  // shares the PCC sourceId (group, project, subproject levels).
+  // Slug is intentionally not updated — it is a stable identifier referenced by FK from
+  // securityInsightsEvaluations and related tables.
+  await db.none(
+    `UPDATE "insightsProjects" ip
+     SET name        = $(name),
+         description = $(description),
+         "logoUrl"   = $(logoUrl),
```
joanagmaia: Can we not update the `logoUrl` for now on Insights projects?
Maybe we can comment it out and leave this comment above:
logoUrl won’t be updated in InsightsProject until we confirm that the format is compatible with the Insights Squared standard.

joanagmaia (dismissed):

services/apps/pcc_sync_worker/src/consumer/pccProjectConsumer.ts
themarolt: done

PR #4054, opened 2026-04-24
feat: infer memberOrganization stint dates from work-email activities (CM-1105)

skwowet:
## Context

When an activity comes in with a verified work email like `jbeulich@suse.com`, we already create a `memberOrganizations` row linking the member to SUSE, but with NULL `dateStart`/`dateEnd`. That causes two problems:

1. A `@suse.com` activity can lose to an unrelated dated enrichment row, because `findAffiliation` treats undated rows as last-resort fallback.
2. The affiliation timeline ignores the row entirely since it has no dates.

We want to use the activity timestamp as evidence that the person was at that company at that moment, and write that into `dateStart`/`dateEnd`. The catch: we can't write on every activity, active maintainers generate hundreds per day, we can't collapse a real multi-stint history like Google -> Apple -> Google into one wrong range, and we can't override user edits or enrichment data.

## How it works

**Hot path** (`data_sink_worker`): When an activity arrives, we buffer `orgId|date` in a Redis set keyed by member, and add the member to a pending-members Redis set. This is done with Redis `MULTI` + `SADD`, so concurrent data-sink workers can safely write the same member/org/date without lost updates. Hundreds of same-day activities collapse to a single set member.

**Cron** (`cron_service`): Every 5 min, the service samples up to 500 pending members, reads each member's buffered set, loads their existing email-domain rows, and walks the buffered dates chronologically applying 4 rules:

1. **Date already inside an existing stint** -> no-op
2. **No dated stint yet for this org** -> stamp the undated placeholder row
3. **Later than `dateEnd`** -> extend forward, with a 30-day debounce and a multi-stint guard. If another org holds a 30+ day stint in the gap, insert a fresh stint instead of bridging.
4. **Earlier than `dateStart`** -> extend backward, same multi-stint guard, no debounce. This is mostly for historical re-ingestion.

After a member is processed successfully, cron atomically removes only the exact Redis set values it read. If new activity arrives while cron is processing, those new values remain buffered for the next run. If cron fails before acking Redis, the dates stay in Redis and are retried. The ack uses a small Lua script with chunked `SREM` calls to avoid large-argument `unpack` limits.

Walking all orgs together in chronological order is what makes multi-stint detection work: by the time the later Google event checks its gap, the Apple stint is already in the working copy and the guard can prevent incorrectly bridging Google -> Apple -> Google into one range.

**`findAffiliation` gets two changes** so email-domain rows start contributing meaningfully:

- **Source priority tier** (`ui > email-domain > enrichment-* > other`) is added to `decidePrimaryOrganizationId`. Once email-domain rows have dates, they beat enrichment on overlaps.
- **Work-email bias**: If the activity payload has a verified work email, `findMemberWorkExperience` pulls in the matching email-domain row as a candidate even when undated. This is the user-visible win that lands immediately: work-email activities resolve to the right org inline, without waiting for the cron to stamp dates.

**Partial index** on `memberOrganizations("memberId") WHERE source = 'email-domain' AND deletedAt IS NULL` backs the cron's per-member fetch so it's a single index seek.

## Cleanup

- Dropped two outdated bin scripts (`map-member-to-org.ts`, `map-tenant-members-to-org.ts`).
- Renamed `memberOrganization.ts` -> `member-organization.ts` to match the folder's kebab-case convention.

---

> [!NOTE]
> **Medium Risk**
> Introduces new background processing that writes/updates `memberOrganizations` date ranges and adjusts affiliation selection logic, which can change organization attribution and timelines if the inference rules are wrong or Redis/cron processing misbehaves.
> 
> **Overview**
> Adds an activity-driven stint-date inference pipeline for `email-domain` member-organization rows: the data-sink buffers `(orgId|date)` evidence in Redis per member, and a new cron job periodically consumes that queue to insert/update dated stints in Postgres.
> 
> Updates affiliation resolution to bias toward verified work-email domains (passing `emailDomain` into `findAffiliation`/`findMemberWorkExperience`) and to prefer higher-priority sources when multiple org stints overlap (`ui` > `email-domain` > `enrichment-*` > other). Manual edits to member organizations now force `source` to `UI` to prevent automated overwrites.
> 
> Adds supporting infra: a partial index for fast per-member `email-domain` lookups, a DAL helper `fetchMemberOrganizationsBySource`, new shared types for stint-change inference, an atomic Redis `ackSetMembers` Lua helper, and removes two obsolete data-sink worker mapping scripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afc5734b99eaaa1da91d0f00c167313731ef721b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/apps/cron_service/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@crowd/cron-service",
   "private": true,
   "scripts": {
-    "start": "SERVICE=cron-service tsx src/main.ts",
+    "start": "SERVICE=cron-service LOG_LEVEL=trace tsx src/main.ts",
```
themarolt: Remove log level trace here.

services/apps/data_sink_worker/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@crowd/data-sink-worker",
   "private": true,
   "scripts": {
-    "start": "SERVICE=data-sink-worker tsx src/main.ts",
+    "start": "SERVICE=data-sink-worker LOG_LEVEL=trace tsx src/main.ts",
```
themarolt: Remove log level trace here.

services/apps/data_sink_worker/src/service/member.service.ts
```diff
@@ -1047,4 +1063,58 @@ export default class MemberService extends LoggerBase {
...
+        this.log.warn(
+          { memberId, organizationId, key },
+          'Corrupt dates buffer value detected during buffering for member organization activity dates.',
+        )
+        return []
+      }
+    }
+
+    // 1. Fetch current dates for this specific organization
+    const existing = await this.redisClient.hGet(key, organizationId)
```
themarolt: The buffer uses non-atomic hGet → JSON parse → push → hSet. Comment acknowledges "occasionally drop a date." With many data_sink_worker pods writing for an active maintainer, this won't be occasional — concurrent same-org writes on the same
  member will lose every-loser-wins.

  Worse, the cron job then does: hGetAll → compute → DB writes → hDel(datesKey, orgIds). Any date buffered between the hGetAll and the hDel is silently destroyed because hDel removes the entire field, not just the values we read. So while the cron
  is running rule evaluation + Postgres writes (potentially seconds), every concurrent activity arriving for that member is at risk.

  Fix: store dates as a Redis set keyed per (member, org) — SADD key:<memberId>:<orgId> <date> is naturally idempotent and concurrency-safe. Use SMEMBERS then SREM with the exact members read (or DEL only after SDIFF confirms no new entries). Or
   use a Lua script for read-and-delete atomicity.

  The "self-healing because future activity will re-populate" reasoning only holds for active orgs. For a member who switches jobs and stops generating activity for the old org, the lost date is the data point.

services/libs/common_services/src/services/common.member.service.ts
```diff
@@ -217,10 +224,28 @@ export class CommonMemberService extends LoggerBase {
...
+          bestRank = rank
+          highestPrioritySourceExperiences = [exp]
+        } else if (rank === bestRank) {
+          highestPrioritySourceExperiences.push(exp)
+        }
+      }
+
+      // Keep only candidates from the highest-priority source tier
+      if (highestPrioritySourceExperiences.length === 1) {
+        return highestPrioritySourceExperiences[0].organizationId
```
themarolt, line 242:
```
if (highestPrioritySourceExperiences.length === 1) {
    return highestPrioritySourceExperiences[0].organizationId
  }
  const memberCounts = await findMemberCountEstimateOfOrganizations(
    this.qx,
    highestPrioritySourceExperiences.map((e) => e.organizationId),
  )
  if (memberCounts[0].memberCount > memberCounts[1].memberCount) 
```

  memberCounts[0]/[1] is hardcoded for 2 candidates. If three+ experiences land in the same priority tier (very plausible: three enrichment-* sources), only the first two are compared and the rest are silently ignored. This bug exists in the
  original code too, but the new tier filter changes which set is passed, so the behavior shifts.

themarolt (changes requested):

services/apps/cron_service/package.json
skwowet: Yeah, added it for debugging during tests. Will remove it after.

services/apps/data_sink_worker/package.json
skwowet: Same here

services/apps/data_sink_worker/src/service/member.service.ts
skwowet: Good point, thanks for catching this. I was thinking mostly about active orgs where a future activity would eventually re-populate the buffer, but you’re right that this breaks down for inactive/old orgs where the lost date may be the only signal.

I updated the Redis model to avoid both races:

- Data-sink now stores buffered dates in a Redis set using `SADD`, with one set per member and values shaped as `organizationId|date`. This makes concurrent writes idempotent and removes the old `hGet -> JSON parse -> hSet` lost-update path.
- Cron now reads the set with `SMEMBERS`, processes the dates, and only after successful DB writes calls a small Lua ack helper that removes exactly the set members it read via `SREM`. It removes the member from the pending queue only if the per-member set is empty after that removal.
- That means any date added while cron is doing inference/Postgres work remains in Redis and will be picked up by a later run.

For cron failure handling, the job still uses `SRANDMEMBER` rather than destructive pop. So if cron fails before the ack, the Redis dates are retained and retried. If DB writes succeed but the ack fails, the same data can be retried, but the next run recomputes against the latest DB state and the writes are intended to be idempotent/recomputable.

services/libs/common_services/src/services/common.member.service.ts
skwowet: Yep, agreed. I updated this to treat the source-priority tier as the candidate set first, then rely on the member-count query ordering to pick a strictly highest winner. 

If there’s no clear winner, it falls back to longest date range within that same priority tier instead of comparing only two hardcoded entries.

services/apps/data_sink_worker/src/service/member.service.ts
themarolt: Ok awesome - lets try it out.

themarolt (dismissed):

PR #4078, opened 2026-05-06
fix: integration results errors with no identity for platform (CM-1111)

themarolt:
> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes ingestion control-flow (some activities are now skipped/marked successful) and relaxes organization creation to allow `undefined` IDs, which could affect attribution and org membership links if assumptions elsewhere rely on prior hard failures.
> 
> **Overview**
> Prevents integration-result failures when an activity/member has no usable identity for the platform. `ActivityService.prepareMemberData` now falls back from missing username to a *verified* same-platform email (and backfills a username identity), treats unverified email as an unrepeatable error, and otherwise skips the activity (warn + mark success) instead of failing the batch.
> 
> Tightens downstream processing to drop payloads whose `activity.username` is still empty after member preparation, avoiding later lookups/upserts on identity-less activities.
> 
> Makes organization linking more tolerant: `OrganizationService.findOrCreate` now returns `string | undefined` (no throw on failure), and `MemberService` only appends org references when an `orgId` is returned while also simplifying the org promise-cache lookup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b71457b88875273ae4c18babd3518c6df6ceeb82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/services/apps/data_sink_worker/src/service/activity.service.ts
+++ b/services/apps/data_sink_worker/src/service/activity.service.ts
@@ -335,8 +335,8 @@ export default class ActivityService extends LoggerBase {
           activity.username = identities[0].value
         } else if (identities.length === 0) {
           // Fall back to same-platform email identity — handles old gerrit records where
-          // only a type:email identity was stored (before parse-member.ts gained the
-          // email-as-username fallback).
+          // only a type:email identity was stored (before the gerrit integration
+          // gained the email-as-username fallback).
           const emailFallback = activity.member.identities.find(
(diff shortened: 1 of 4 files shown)


joanagmaia (approved):
LGTM @mbani01 probably worth investigating a bit on why we don't get emails for some activities. Will create a ticket with an example so we can look into it

PR #4088, opened 2026-05-07
fix: prevent stale activityRelations in affiliations refresh (CM-1132)

skwowet:
## Summary

When a member had `memberSegmentAffiliations` (MSA) rows, the affiliation timeline builder mixed them together with `memberOrganizations` into a single timeline. `selectPrimaryWorkExperience` would pick the MSA as the primary org for any overlapping date range (MSAs have highest precedence), producing a `TimelineItem` with a `segmentId`.

`processAffiliationActivities` then applied `WHERE segmentId = X` for that item, so activities in other segments were never updated. This left stale `organizationId` values in `activityRelations` for the unprocessed segments.

## What changed

The timeline is now split into two passes:

1. **Base timeline**
   Built from `memberOrganizations` only, without a `segmentId`, and applies globally across all segments. When MSAs exist, each item is flagged with `skipManualAffiliationSegments`.

2. **Manual timeline**
   Built per MSA group (grouped by `segmentId`), with each item scoped to its segment. Gap items (null org) are discarded since the base timeline already covers those date ranges.

In `processAffiliationActivities`, the `skipManualAffiliationSegments` flag adds a `NOT EXISTS (SELECT 1 FROM memberSegmentAffiliations ...)` condition so the base pass skips activity rows already covered by an MSA. Those rows are then handled by the manual pass.

This ensures every `activityRelations` row is updated exactly once, with no double updates and no missed rows.

---

> [!NOTE]
> **Medium Risk**
> Touches timeline generation and the SQL update criteria used to rewrite `activityRelations.organizationId`, so mistakes could mis-assign orgs or skip updates across segments despite being a targeted fix.
> 
> **Overview**
> Fixes affiliation refresh so `activityRelations` updates don’t get skipped for members with `memberSegmentAffiliations`.
> 
> The timeline build is split into a **global base pass** (from `memberOrganizations`) and a **per-segment manual pass** (from MSAs). Base items can be flagged with `skipManualAffiliationSegments`, and `processAffiliationActivities` now adds a `NOT EXISTS` guard so the base pass won’t overwrite rows covered by MSAs; the manual pass then updates those segment-scoped rows. Also adds an option to omit fallback ranges when building manual timelines and removes redundant try/catch wrappers in the enrichment worker activity functions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 826182307024a493f494ec9ff6b6cccea9e3c758. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/libs/data-access-layer/src/member-organization-affiliation/index.ts
```diff
@@ -327,31 +358,47 @@ async function processAffiliationActivities(
...
+  if (affiliation.skipManualAffiliationSegments) {
+    conditions.push(`
+      NOT EXISTS (
+        SELECT 1
+        FROM "memberSegmentAffiliations" msa
+        WHERE msa."memberId" = $(memberId)
+          AND msa."segmentId" = ar."segmentId"
+          AND msa."organizationId" IS NOT NULL
+          AND (msa."dateStart" IS NULL OR ar."timestamp" >= msa."dateStart"::date)
+          AND (msa."dateEnd" IS NULL OR ar."timestamp" < msa."dateEnd"::date + interval '1 day')
```
epipav, line 397: Why do we have `dateEnd + interval '1 day'` here?

epipav (commented):
looks good 👍  added one comment

services/libs/data-access-layer/src/member-organization-affiliation/index.ts
skwowet: `dateEnd` is a date but `ar."timestamp"` is a full timestamp. 

Casting `dateEnd::date` gives midnight (`2023-09-01 00:00:00`), which would exclude all activities that happened during that day. The `+ interval '1 day'` shifts the boundary to the next midnight, making it inclusive of the entire dateEnd day.

PR #4154, opened 2026-05-29
chore: optimize query members to merge

joanagmaia:


joanagmaia:
This pull request is no longer needed. Optimizations on the suggestions endpoint can be done later on.

PR #4155, opened 2026-05-29
docs: oss packages criticality (CM-1207)

joanagmaia:
This pull request significantly expands and finalizes the methodology for scoring package criticality in the `oss-packages` domain, replacing the previous placeholder formula with a detailed, defensible approach. It introduces a new section describing the signals, formula, tier allocation policy, and manual override system for identifying critical packages across ecosystems. The documentation is updated to reflect these decisions and clarify open questions.

**Criticality scoring methodology and related changes:**

* Added a comprehensive "Criticality scoring methodology" section detailing the signals (including new graph-derived metrics like transitive dependent count and PageRank centrality), the percentile-rank-based scoring formula, and the rationale for weighting each input. This replaces the previous placeholder and defines a robust, tunable approach for ranking package criticality.
* Defined the per-ecosystem tier allocation policy (floor + ceiling + judgment) for critical packages, including example tier budgets and the reasoning behind them.
* Introduced a "spotlight override" mechanism using a new `package_criticality_spotlight` table, allowing manual promotion of packages to critical status with required rationale and auditability.
* Documented the implementation approach for graph-based signals, including in-memory computation versus possible ingestion from deps.dev, and outlined the worker layout for the new criticality sub-worker.

**Documentation and tracking updates:**

* Updated the summary table and open questions to reflect the finalized criticality scoring methodology, removal of the placeholder formula, and new open questions regarding deps.dev coverage for graph signals. [[1]](diffhunk://#diff-3872dba0cefa1514b68ed6df84e129c0e4cbe8d2e2c5b9170588faa99e3f0126L17-R18) [[2]](diffhunk://#diff-3872dba0cefa1514b68ed6df84e129c0e4cbe8d2e2c5b9170588faa99e3f0126R570)
* Added a changelog entry summarizing the addition of the criticality methodology and the folding of related standalone ADRs into the main record.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to a living ADR; no runtime, schema, or auth changes in this diff.
> 
> **Overview**
> This PR **replaces the placeholder criticality formula** in `docs/adr/0001-oss-packages-design-decisions.md` with a full **§Criticality scoring methodology** and updates the scope table, universe section cross-links, open questions, and changelog.
> 
> The new section defines **five ranking signals** on `packages_universe` (adding `transitive_dependent_count` and `centrality_score`), a **per-ecosystem percentile-rank + weighted blend** formula in `[0, 1]` with tunable call-time weights and suggested starting weights, and rationale for **percentile-rank vs min-max** and **per-ecosystem** normalization. It also documents **floor + ceiling tier budgets** via `critical_top_n_by_ecosystem`, an auditable **`package_criticality_spotlight`** override applied after ranking, and a planned **`packages_worker/src/criticality/`** Temporal sub-worker (weekly, per-ecosystem) with optional **in-memory PageRank/BFS** vs ingesting transitive counts from deps.dev if available. Status is **proposed** (weights and deps.dev verification still open).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d6ed9c2d6acaa6e03f7c59ecb6febf8f84076b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/adr/0001-oss-packages-design-decisions.md
```diff
@@ -82,14 +83,140 @@ Tier 2 enriches a critical slice of the npm and Maven ecosystems — not the ful
...
+
+**Before committing to this implementation, confirm whether deps.dev already provides these signals so we can ingest instead of compute:**
+
+- **Transitive dependent count** — `DependentsLatest` is the table we already source `dependent_packages_count` from. Verify whether its dependent counts are direct-only or include indirect dependents. If indirect counts are included, the column can be sourced in the existing universe-import job (consistent with how the other dependent counts are already populated) and the in-memory transitive computation is unnecessary.
+- **Centrality / importance score** — deps.dev does not appear to expose a PageRank-style score in its current schema.
```
mbani01, line 176: The dependents_count from deps.dev should already include the transitive ones implicitly (as long as we don't limit `MinimumDepth`).

From [docs](https://blog.deps.dev/enumerating-dependents/#:~:text=Sample%20queries,All%20dependent%20versions):
We compute a full set of transitive dependencies for each version of each package, and we call this the “dependency graph”. This data is available on our [web site](https://deps.dev/), [API](https://docs.deps.dev/api/v3alpha/#getdependencies), and [BigQuery](https://docs.deps.dev/bigquery/v1/#dependencygraphedges) dataset. We also compute the inverse of these dependency graphs, providing the full set of versions that depend on any given version, and we call these “dependents”.

I managed to get the breakdown using API (example below), while @themarolt is checking changes required to include this on bigquery export.

curl 'https://api.deps.dev/v3alpha/systems/maven/packages/net.gleamynode%3Anetty2/versions/1.9.2:dependents'

{
    "dependentCount": 23,
    "directDependentCount": 4,
    "**indirectDependentCount**": 20
}

So I believe there is no need to compute or store this in db, if we'll be using `dependent_packages_count` as the source of truth

docs/adr/0001-oss-packages-design-decisions.md
```diff
@@ -82,14 +83,140 @@ Tier 2 enriches a critical slice of the npm and Maven ecosystems — not the ful
...
+
+Five signals, all stored on `packages_universe`:
+
+| Signal                       | Existing? | Source                                                                  |
+| ---------------------------- | --------- | ----------------------------------------------------------------------- |
+| `downloads_last_30d`         | yes       | weekly downloads ingestion (registry APIs)                              |
+| `dependent_packages_count`   | yes       | deps.dev `DependentsLatest`                                             |
+| `dependent_repos_count`      | yes       | derived in Postgres from `package_repos`                                |
+| `transitive_dependent_count` | **new**   | computed in the criticality sub-worker (see Implementation note below) |
+| `centrality_score`           | **new**   | computed in the criticality sub-worker (PageRank, see below)            |
```
mbani01, line 110: The existing implementation covers only the direct dependency graph (transitive dependency relations aren't tracked). Will centrality_score be based on direct deps only?

themarolt (approved):

docs/adr/0001-oss-packages-design-decisions.md
joanagmaia: No, it should be based on both.

docs/adr/0001-oss-packages-design-decisions.md
joanagmaia: we should be able to store the count for each one of the data points. If the dependentCount gives us total, we can still store it. But if that's the case we would also need to store a new data point for directDependentCont and transitiveDependentCount. On the formula the goal would be to give different weights to each, so we shouldn't aggregate this into one data point.

mbani01 (approved):

PR #4170, opened 2026-06-04
fix: soft delete member segment affiliations for blocked orgs

skwowet:
## Summary

This PR adds soft-delete support for `memberSegmentAffiliations` so manual segment affiliations can be preserved while being excluded from active affiliation logic.

When an organization affiliation is blocked, related member segment affiliation rows are now soft-deleted as well. This prevents manually created MSAs from keeping activities affiliated to blocked organizations during recalculation, while still retaining the original data through `deletedAt`.

## Changes

- Add `deletedAt` to `memberSegmentAffiliations`
- Make MSA deletion soft by default, while keeping hard delete available for flows where it is still appropriate
- Filter soft-deleted MSAs from read paths, project affiliation APIs, OpenSearch, and affiliation recalculation
- Soft-delete related MSAs when org affiliation blocking or member-org blocking is applied
- Keep soft-deleted MSAs with the member during merge/unmerge moves so historical rows are not left behind
- Clean up older member segment affiliation code paths to use the shared MSA helpers consistently

## Notes

- Unblocking an organization does not automatically restore previously soft-deleted MSAs
- Replacing a member’s project affiliations soft-deletes the previous active rows before inserting the new set


themarolt (dismissed):

services/libs/data-access-layer/src/members/base.ts
```diff
@@ -589,14 +589,18 @@ export async function moveAffiliationsBetweenMembers(
...
+      update "memberSegmentAffiliations"
+      set "memberId" = $(toMemberId)
+      where
+        "memberId" = $(fromMemberId)
+        and "deletedAt" is null
+    `,
+    {
+      fromMemberId,
+      toMemberId,
+    },
```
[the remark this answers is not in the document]
skwowet: Fixed ✅

backend/src/services/organizationService.ts
```diff
@@ -1074,6 +1075,9 @@ export default class OrganizationService extends LoggerBase {
         data.isAffiliationBlocked !== existingOrg.isAffiliationBlocked
       ) {
         await applyOrganizationAffiliationPolicyToMembers(qx, record.id, !data.isAffiliationBlocked)
+        if (data.isAffiliationBlocked) {
+          await deleteMemberSegmentAffiliations(qx, { organizationId: record.id })
+        }
```
ulemons, line 1080: i see different calls here, dont we need a tx or something ?

ulemons (dismissed):

backend/src/services/organizationService.ts
skwowet: Good catch — we’re already in one. `update()` opens a transactional `repoOptions`, and `qx` from `getQueryExecutor` is bound to that same Sequelize tx, so the org update, override batch, and MSA soft-delete all commit or roll back together. Recalc workflow still fires after commit on purpose.

PR #4188, opened 2026-06-10
fix: prefer company over uni on overlap during affiliation (CM-1216)

skwowet:
## Summary

Affiliation logic now treats the email on an activity as the primary signal for determining which organization should be credited for that activity. When an activity does not contain an organization email, overlapping university and company affiliations are resolved in favor of the company.

This behavior now applies consistently during both ingest (`data_sink_worker`) and bulk refresh (`refreshMemberOrganizationAffiliations`).

## Context

Member activities can be affiliated to an organization based on work history (`memberOrganizations`), manual segment rules, and other fallback logic. That generally worked well for simple timelines, but broke down in a couple of cases:

* A member had overlapping university and company roles (e.g. student + internship).
* An activity contained an explicit organization email (`user@company.com`, `user@school.edu`), but date-based affiliation logic resolved the activity to a different organization.

The email used on an activity is first-party evidence of which organization the activity was performed on behalf of. It should take precedence over enrichment-derived timelines and generic tie-breakers (stint length, org size, source rank), unless a manual affiliation override has been configured.

### Ingest (`findAffiliation`)

Affiliation resolution now follows this order:

1. Manual segment affiliations (unchanged).
2. Activity email domain — if the activity username contains an organization email, affiliate to the `memberOrganization` whose organization has a matching verified `primary-domain`.
3. Date-based matching — for activities without an organization email (plain handles, public inbox domains), affiliate using the organization active at that timestamp. When a company and university overlap, the company is preferred.
4. Existing undated fallbacks (unchanged).

Only the email present on the activity username is considered. Member profile emails are not used as a substitute when the activity itself does not contain an email.

### Bulk refresh (`prepareMemberOrganizationAffiliationTimeline`)

Bulk refresh now mirrors the same behavior used during ingest.

To avoid activities being processed twice, refresh runs in two passes:

| Pass | Activities | Logic |
|------|------------|--------|
| Email-based | Username matches a member organization's verified domain | Affiliate directly to that organization |
| Timeline-based | No email, or no matching member organization | Use date-based affiliation logic, preferring companies over universities when overlaps exist |

Manual segment affiliations continue to take precedence through the existing segment-scoped timeline and `skipManualAffiliationSegments` guards.

## Expected behavior

| Activity signal | Result |
|-----------------|--------|
| Username contains an organization email (`@company.com`, `@school.edu`) | Affiliate to the organization with the matching verified domain |
| No organization email, company + university overlap | Company wins |
| Manual segment affiliation | Manual rule wins for that segment |


services/libs/data-access-layer/src/organizations/identities.ts
```diff
@@ -144,6 +144,81 @@ export async function upsertOrgIdentities(
...
+
+  // 2. Single pass: Check the heuristic and categorize right here
+  for (const candidate of candidates) {
+    const org = orgsById.get(candidate.organizationId)
+
+    // If we find the org, check if it's a university based on name or .edu domain
+    const isUniversity = org
+      ? (org.displayName?.includes('university') ?? false) ||
+        org.domains.some((d) => d.endsWith('.edu'))
+      : false
```
ulemons, line 207: maybe we can extend this to some other domain, im thinking my university for example (unibo.it) or "mit" or something, should we have a separate list for this that we can enrich later ?

ulemons (dismissed):

services/libs/data-access-layer/src/organizations/identities.ts
skwowet: Yeah, good idea. I was thinking the same thing, but kept this PR intentionally small with the current heuristic (`.edu` domain or org name containing “university”).

We can definitely add a curated list of university domains/patterns, similar to how we maintain excluded or blacklisted domains today, and expand it as we encounter more real-world cases.

PR #4225, opened 2026-06-17
fix: dedupe overlapping email-domain member organizations (CM-1189)

skwowet:
## Summary

Members can end up with duplicate member organization rows when an email-domain inferred stint overlaps a non-email-domain row for the same member and organization. That shows up as duplicate work history in both the public API and the internal member organization UI.

This change treats those overlapping rows as one visible member organization on read, while keeping the underlying email-domain rows in sync on write. It also updates stint inference so soft-deleted dated email-domain rows are not recreated for the same date ranges.

## Example

Before, a member with a UI row and an email-domain row for the same org and overlapping dates would show two entries:

```
id: abc  org: Google  dateStart: 2021-01-01  dateEnd: null       source: ui
id: xyz  org: Google  dateStart: 2020-06-01  dateEnd: 2023-01-01 source: email-domain
```

After, they are grouped into one:

```
id: abc  org: Google  dateStart: 2020-06-01  dateEnd: null  source: ui,email-domain
```

The display row is always the non-email-domain one (`ui` > enrichment by source rank). The date range is the union of both rows. When there is no overlap, both rows continue to show separately.

## Changes

- Added grouping logic to merge overlapping email-domain member organizations into the best non-email-domain display row, using source priority, combined sources, and a unioned date range.
- Updated public work experience reads to return grouped member organizations instead of separate overlapping rows.
- Updated public update, delete, and verify flows to fan out shared field changes to overlapping email-domain rows, while leaving date edits on the visible row only.
- Updated internal member organization list, update, delete, and affiliation override flows to use the same grouped read and write behavior.
- Extended source-based member organization fetching to optionally include soft-deleted rows for stint inference.
- Updated stint inference to normalize dates once, treat epoch-like values as missing, and skip recreating email-domain stints inside soft-deleted dated ranges.


backend/src/utils/mapper.ts
```diff
@@ -1,4 +1,127 @@
...
+          getMemberOrganizationSourceRank(a.source) - getMemberOrganizationSourceRank(b.source)
+        if (rankDiff !== 0) {
+          return rankDiff
+        }
+
+        return (a.id ?? '').localeCompare(b.id ?? '')
+      })[0]
+
+      if (displayRow.id) {
+        hiddenEmailDomainIds.add(displayRow.id)
```
ulemons: Is it ok to have this displayRow.id here ? maybe it should be something like emailDomainRow.id ?

backend/src/services/member/memberAffiliationsService.ts
```diff
@@ -102,11 +108,42 @@ export default class MemberAffiliationsService extends LoggerBase {
...
+        memberOrganizationId,
+      })),
     )
 
+    if (data.allowAffiliation === false && memberOrg?.organizationId) {
+      await deleteMemberSegmentAffiliations(qx, {
+        memberId: data.memberId,
+        organizationId: memberOrg.organizationId,
+      })
+    }
```
ulemons, line 139: is the `deleteMemberSegmentAffiliations` deleting the entire org history, not to the specific work experience being overridden? what if the member has two non-overlapping stints at the same org and `allowAffiliation` is set to `false` on one, the other loses its segment affiliations too ? is this a real use case ? maybe not

ulemons (dismissed):

backend/src/utils/mapper.ts
skwowet: Good catch — you're right, it should be `emailDomainRow.id`.

`hiddenEmailDomainIds` is the set of email-domain rows we filter out of the raw list. `displayRow.id` is only the grouping key in `displayGroups` (the visible non-email-domain row we merge into).

Using `displayRow.id` there would hide the display row instead of the email-domain row. I'll fix that.

backend/src/services/member/memberAffiliationsService.ts
skwowet: Yeah, when `allowAffiliation` is false we delete segment affiliations for the whole member+org, not just that one work experience stint.

That behavior was already there before this PR — we didn't change it, just kept the same logic when we refactored the override path. 

The two-stints-at-same-org case is possible in theory, but I haven't seen it come up.

backend/src/utils/mapper.ts
skwowet: Fixed ✅

Thanks!

PR #4239, opened 2026-06-19
fix: case insensitive channel matching in activityRelations enrich pipes (IN-1185)

joanagmaia:
## Problem

Activities from GitHub have their `channel` field set directly from the GitHub API response, which preserves the original repository owner casing (e.g. `https://github.com/NousResearch/hermes-agent`).

The `repositories` Tinybird datasource, however, stores URLs in lowercase (e.g. `https://github.com/nousresearch/hermes-agent`). This means that when the `activityRelations_bucket_clean_enrich_copy_pipe_*` pipes filter activities using:

```sql
AND (channel, segmentId) IN (SELECT channel, segmentId FROM repos_to_channels)
```

the tuple comparison is case-sensitive in ClickHouse, so activities whose channel URL has a different casing than the stored repository URL are silently dropped from `activityRelations_deduplicated_cleaned_bucket_*_ds`.

Concretely: querying `activityRelations` for a given segment and type returns rows, but querying `activityRelations_deduplicated_cleaned_bucket_union` for the same filters returns 0 — because the channel casing mismatch causes the filter to exclude all matching activities.

## Fix

Apply `lower()` to both sides of the channel comparison in all 10 `activityRelations_bucket_clean_enrich_copy_pipe_*.pipe` files:

```sql
-- before
AND (channel, segmentId) IN (SELECT channel, segmentId FROM repos_to_channels)

-- after
AND (lower(channel), segmentId) IN (SELECT lower(channel), segmentId FROM repos_to_channels)
```

This is safe for all platforms (GitHub, GitLab, Gerrit) since `lower()` is idempotent when the URL is already lowercase. After deploying, the next scheduled copy run (every 10 minutes) will backfill previously-dropped activities.

---

> [!NOTE]
> **Medium Risk**
> Changes analytics ETL filtering for all git platforms; wrong normalization could include or exclude activities until the next COPY replace run.
> 
> **Overview**
> Fixes git-related activities being **dropped** from `activityRelations_deduplicated_cleaned_bucket_*` when activity `channel` URLs keep GitHub API casing but `repos_to_channels` stores lowercase repository URLs.
> 
> In all ten `activityRelations_bucket_clean_enrich_copy_pipe_*.pipe` files, the git platform filter now compares **`lower(channel)`** on both sides of the `(channel, segmentId)` tuple `IN` subquery against `repos_to_channels`, so casing no longer excludes valid rows from the scheduled COPY into the cleaned bucket datasources.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6b9680afe9650e765a327c131bc4ec931be6868. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gaspergrom (approved):

gaspergrom:
Lgtm

PR #4240, opened 2026-06-19
refactor: move segment subproject expansion out of middleware

skwowet:
## Summary

CM-189 expanded every request’s segments to subprojects inside `segmentMiddleware`, so `req.currentSegments` often held hundreds of IDs before handlers ran. This PR reverts that middleware behavior and moves subproject resolution into the DAL, applied only where the data model requires it.

Middleware goes back to `findInIds` — `currentSegments` matches what the client sent (usually a project group ID). Subproject expansion happens at specific call sites via `getSegmentSubprojects` / `getSegmentSubprojectIds`. Rollup-based paths (member/org query, bot suggestions) keep using the request segment ID unchanged.

## Changes

**Middleware**
- Remove `resolveToLeafSegments` and all leaf-expansion logic
- Resolve segments with `findInIds` only; keep `toStringArray` for query/body parsing

**DAL & segment layer**
- Add `getSegmentSubprojects` and `getSegmentSubprojectIds` to `data-access-layer`
- Remove duplicate `getSegmentSubprojects` from `segmentRepository`
- Point `segmentService.getSegmentSubprojects` at the DAL

**Activity**
- Expand segments in `activityService` for query, types, and channels
- Remove `getTenantSubprojects` and the “load all tenant subprojects” fallback
- Activity API handlers use `currentSegments` from middleware instead of `req.query.segments`

**Member & organization repositories**
- Expand to subproject IDs for junction writes, excludes, autocomplete, activity counts, and merge suggestions

**Integrations**
- Expand segments in `update` and `findAndCountAll` so `integration/query` works when the frontend sends only a project group ID (CM-189 `ActivityTypeFilter` path)

**Cleanup**
- Drop unused middleware imports, logging, and `activityTypes` cloning in middleware
- Fix spurious `await` on sync `getTenantActivityTypes` in `settingsRepository`
- Update `auth-axios` comment to match the new middleware behavior


ulemons (dismissed):

ulemons:
Overall looks good to me. 
Before merging I would check 2 things:
1. is there any other service referring to subprojectId which now will get projectGroupId ? (e.g. `memberService`, `integrationSerice`.. )
2. Before the expand was made one time in the middleware, are we sure we are not calling the getSegmentSubprojectIds multiple times now ? (e.g. create -> calls to update or something )

skwowet:
Checked both:
1. Audited `getSegmentIds` / `currentSegments` usages — subproject-scoped paths (activities, integrations query, member/org junction tables, merge suggestions) use `getSegmentSubprojectIds`. Rollup paths (`memberService.query`, `orgService.query`, bot suggestions) intentionally use project group ID. `memberService` import uses agg rollup, not junction.
2. Expansion is now per call site, not per request. Main paths expand at most once per handler and reuse the result (e.g. member create/update). No double expand from middleware + repo.

PR #4261, opened 2026-06-25
perf: use `hasMore` instead of live count on merge suggestion lists

skwowet:
## Summary

List requests for merge suggestions were still slow on large project groups because every page load ran a live `COUNT(*)` in addition to the list query. This removes that bundled count: the API fetches one extra row to determine `hasMore`, and the UI paginates from that instead of `res.count`. Explicit `countOnly` calls are unchanged.

## Changes

- Stop calling `getTotalCount()` on list responses in `memberRepository` and `organizationRepository`; return `hasMore` from a `limit + 1` query instead of `count`
- Member and organization merge suggestions pages use `hasMore` for the load-more button instead of comparing `total` to loaded rows
- Data quality member/org merge suggestion lists use the same `hasMore` pattern for load more
- Member and organization merge suggestion dialogs use `hasMore` for next-page navigation and `hasSuggestion` for empty/actions state, replacing total-count-based prev/next and “X of Y” labels
- After merge or ignore in dialogs, refetch at the current offset and step back one page when that offset is empty

### DIFF
--- a/backend/src/database/repositories/memberRepository.ts
+++ b/backend/src/database/repositories/memberRepository.ts
@@ -410,6 +410,9 @@ class MemberRepository {
       return { count: await getTotalCount() }
     }
 
+    const pageLimit = args.limit
+    const queryLimit = pageLimit + 1
+
     const mems = await options.database.sequelize.query(
       `
         SELECT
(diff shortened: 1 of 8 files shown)


joanagmaia (approved):
LGTM

PR #4280, opened 2026-06-30
feat: discover and parse project-maintained affiliations (CM-361)

skwowet:
## Summary

This PR adds processing for project-maintained contributor-to-employer files inside `git_integration`.

Many LF projects publish who works for which company in repo files (gitdm, `.organizationmap`, SIG YAML, `INFO.yaml`, etc.). Formats and filenames vary, so we reuse the existing clone pipeline: find the file on disk, parse it with AI into a normalized snapshot, track per-repo state, and resolve rows against CDP members and organizations.

The worker runs this on the first clone batch, alongside maintainer processing. Each repo gets one registry row (`git.repoAffiliationRegistry`, #4279) plus a `serviceExecutions` record with operation type `Affiliation`.

**Run flow**

```
Interval elapsed?
  no  → skip run
  yes ↓

Resolve file path
  saved path still on disk? → use it
  else discover:
    list text-like files at repo root → AI picks from candidates (path + 400 char preview per file)
  ↓

Parse
  file hash unchanged and snapshot valid? → reuse snapshot (no AI)
  else → AI extract flat rows, group by contributor into snapshot
  ↓

Fork? → drop stints already on parent repo's cached snapshot

Apply (dry run in this PR)
  for each stint: lookup member + org, run guards, build MO/MSA inserts
  actual INSERT calls are commented out
  ↓

Update registry (filePath, fileHash, status, snapshot) + serviceExecutions
```

**Parsed snapshot shape**

Each snapshot entry is one contributor with their organization stints over time:

- **Contributor:** email or GitHub username required (email preferred for grouping and lookup). Name is optional.
- **Organizations:** each stint has domain (required for CDP lookup), optional name, optional `dateStart`/`dateEnd`, and `isUnaffiliated` for independent/unaffiliated rows (normalized in code to `individual-noaccount.com`). AI returns flat contributor–org rows; code groups by contributor and dedupes stints. Rows missing email/github or domain are dropped during normalization.

**Registry status:** `success` (including zero-row parses), `not_found`, `unusable` (parsed but nothing applyable; stable until file content changes), `error` (unexpected parse failure; retry on next interval).

**Apply rules (implemented, writes disabled for now)**

- Lookup only: no find-or-create for members or orgs. Unresolved rows are skipped and retried on the next run using the same snapshot.
- Member identity: email first (git username identity), else GitHub username.
- Org identity: verified `primary-domain` only (no display-name lookup).
- Guards: skip MO/MSA insert when the same member + org + date range already exists, or an undated/open-ended row already covers an undated insert (MSA scoped to the repo segment).
- MO source would be `project-registry` when inserts are enabled.

Repos without a segment skip apply. Unexpected parse failures set registry status to `error`, clear `fileHash`, and keep the previous snapshot so the next interval run retries parsing.

## Changes

**AffiliationService** (`services/affiliation/affiliation_service.py`)
- Root-level text file discovery; AI file picker on all candidates (no known-filename list). Batches candidates (20 per call) with short file previews.
- **Precision-first prompts:** discovery prefers `not_found` over a wrong file; rejects plain name/email lists (`.mailmap`, `AUTHORS`, `CONTRIBUTORS`), governance without employer info (`OWNERS`, bare `MAINTAINERS`), and scripts/config. Allows gitdm-style files where the file itself defines explicit domain/email-pattern affiliation rules.
- **Extraction:** only emit rows when the file explicitly states an org or explicit unaffiliation. No org/domain inference from raw emails or usernames; domain may be inferred from a stated org name when confident. `isUnaffiliated` only when the file explicitly marks it (AI sets name/domain null; code maps to Individual).
- AI extraction returns flat rows with optional dates and unaffiliated markers; `group_parse_rows` merges into contributor + `organizations[]` snapshot shape. Malformed parse responses retry once before `unusable`.
- AI extraction with chunked parsing for files over 5k chars.
- SHA-gated snapshot caching to avoid re-parsing unchanged files. Invalid/old snapshot shape fails validation and forces re-parse.
- Separate update interval (24h when a file is found) and retry interval (30d when not found).
- Fork repos: filter stints already present on parent repo snapshot before apply.
- Full apply path: bulk identity resolution, dedupe, stint-level guard checks, insert list building. MO/MSA `executemany` calls commented out with a TODO.

**Worker wiring** (`repository_worker.py`, `server.py`)
- `AffiliationService` registered and invoked on `is_first_batch` after maintainer processing.

**Database layer** (`crud.py`)
- Registry get/upsert and snapshot JSON serialize/deserialize via Pydantic.
- Bulk member/org identity lookups (single query per entity type, index-aligned results).
- Fetch existing MO and segment-scoped MSA for guards.
- Insert helpers for MO (`ON CONFLICT DO NOTHING` aligned to partial unique indexes) and MSA.

**Shared infrastructure**
- Bedrock client moved from `maintainer/bedrock.py` to `services/llm/bedrock.py`; maintainer import updated.
- New Pydantic models (`affiliation_info.py`), error types, `AffiliationRegistryStatus`, `OperationType.AFFILIATION`, and `AFFILIATION_*` settings in `settings.py` / `conftest.py`.


services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +526,289 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+        snapshot = snapshot["affiliations"]
+    try:
+        return _AFFILIATION_SNAPSHOT_ADAPTER.validate_python(snapshot)
+    except ValidationError as error:
+        logger.warning(f"Invalid affiliation snapshot in registry, will re-parse: {error}")
+        return None
+
+
+def dump_affiliation_snapshot(affiliations: list[AffiliationInfoItem]) -> list[dict]:
+    return [item.model_dump() for item in affiliations]
```
mbani01: These are mainly used for parsing and validation logic. Ideally, we should keep crud.py focused solely on database-related operations.

Let's create a dedicated model for `repoAffiliationRegistry` similar to [the Repository model](https://github.com/linuxfoundation/crowd.dev/blob/main/services/apps/git_integration/src/crowdgit/models/repository.py) so that we can use the model directly instead of handling multiple fields separately in validation, function arguments, and return types.

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +526,289 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+    if snapshot is not None:
+        snapshot = parse_affiliation_snapshot(snapshot)
+
+    return {
+        "file_path": row.get("filePath"),
+        "file_hash": row.get("fileHash"),
+        "status": row.get("status"),
+        "snapshot": snapshot,
+        "last_run_at": row.get("lastRunAt"),
+    }
```
mbani01: Once the [repoAffiliationRegistry model](https://github.com/linuxfoundation/crowd.dev/pull/4280/changes#r3498620136) is defined, let it handle this ([example](https://github.com/linuxfoundation/crowd.dev/blob/31dc4470edf4de2293a59ea5811102b1e42a423b/services/apps/git_integration/src/crowdgit/database/crud.py#L61)):

```suggestion
    return repoAffiliationRegistry.from_db(dict(result)) if result else None
```

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +526,289 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+
+
+async def upsert_repo_affiliation_registry(
+    repo_id: str,
+    *,
+    file_path: str | None,
+    file_hash: str | None,
+    status: str,
+    snapshot: list[AffiliationInfoItem] | None,
```
mbani01: We can use [repoAffiliationRegistry](https://github.com/linuxfoundation/crowd.dev/pull/4280/changes#r3498620136) here as well:

```suggestion
async def upsert_repo_affiliation_registry(repoAffiliationRegistry)
```

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +526,289 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+        [
+            (
+                row["member_id"],
+                row["organization_id"],
+                row.get("source", "project-registry"),
+            )
+            for row in rows
+        ],
+    )
+    return len(rows)
```
mbani01: This returns the number of rows ****attempted**, not the number actually inserted** - so it might be misleading as we're using on conflicts do nothing, if we really need the count of inserted records let's use `RETURNING` on the query

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +526,289 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+            (
+                row["member_id"],
+                row["segment_id"],
+                row["organization_id"],
+                row.get("verified", False),
+            )
+            for row in rows
+        ],
+    )
+    return len(rows)
```
mbani01: Same as https://github.com/linuxfoundation/crowd.dev/pull/4280/changes#r3498854974

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,887 @@
...
+        execution_status = ExecutionStatus.SUCCESS
+        error_code = None
+        error_message = None
+        ai_cost = 0.0
+        latest_file_path: str | None = None
+        latest_file_hash: str | None = None
+        registry = await get_repo_affiliation_registry(repository.id)
+
+        try:
+            has_interval_elapsed, _ = await self.check_if_interval_elapsed(registry)
```
mbani01: I see `remaining_hours` returned by `check_if_interval_elapsed` isn't being used.
We can add it to logs if useful, otherwise let's remove the extra return and keep only  `has_interval_elapsed`

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,887 @@
...
+        if picked_path and await aiofiles.os.path.isfile(os.path.join(repo_path, picked_path)):
+            return picked_path, ai_cost
+
+        return None, ai_cost
+
+    async def resolve_affiliation_file(
+        self,
+        repo_path: str,
+        saved_file_path: str | None,
+        repo_url: str = "",
```
mbani01: I don't think there's a case when repo_url is null, maybe no need for the default value

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,887 @@
...
+                if not await aiofiles.os.path.isfile(full_path):
+                    self.logger.debug(f"AI picked path not on disk, skipping: {picked_path!r}")
+                    continue
+                self.logger.info(f"Affiliation file: {picked_path} (AI)")
+                return picked_path, total_cost
+
+        return None, total_cost
+
+    async def discover_affiliation_file(
+        self, repo_path: str, repo_url: str = ""
```
mbani01: same here for repo_url: https://github.com/linuxfoundation/crowd.dev/pull/4280/changes#r3498920560

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,887 @@
...
+                cwd=repo_path,
+            )
+        except CommandExecutionError:
+            return []
+        except FileNotFoundError:
+            self.logger.warning("Ripgrep not found, known filename search is unavailable")
+            return []
+        except Exception as e:
+            self.logger.warning(f"Known filename search failed for {known_name!r}: {repr(e)}")
+            return []
```
mbani01: The first except might be silently failing, let's remove it so it's handled by the last one and get logs

```suggestion
        except FileNotFoundError:
            self.logger.warning("Ripgrep not found, known filename search is unavailable")
            return []
        except Exception as e:
            self.logger.warning(f"Known filename search failed for {known_name!r}: {repr(e)}")
            return []
```

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,887 @@
...
+        for line in output.strip().split("\n"):
+            line = line.strip()
+            if not line:
+                continue
+            if line.startswith("./"):
+                line = line[2:]
+            if self.path_matches_known_name(line, known_name):
+                matches.append(line)
+
+        return sorted(matches)
```
mbani01: I see we'll be sorting the final matches in `find_known_file_matches`, so we skip sorting here 
```suggestion
        return matches
```

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,887 @@
...
+                )
+        else:
+            candidates = await self.list_root_text_files(repo_path)
+            root_files_only = True
+            self.logger.debug(
+                f"No known filename matches, checking {len(candidates)} repo root files with AI"
+            )
+
+        if not candidates:
+            return None, ai_cost
```
mbani01: Is there a specific reason why we don't perform the is_text_file_path check when `len(matches) == 1` ?

Also, we can perform all the checks (name, extension, root level) in one go, since `rg` supports it (you can check the example in [maintainers](https://github.com/linuxfoundation/crowd.dev/blob/d670a9adae526dc7d6b4616acab68ec84e1f1a0f/services/apps/git_integration/src/crowdgit/services/maintainer/maintainer_service.py#L559))

services/apps/git_integration/src/crowdgit/database/crud.py
skwowet: Done. Added `RepoAffiliationRegistry` to `affiliation_info.py` (same file as the parse models) with `from_db()` and `snapshot_for_db()`. Snapshot parse/validation moved there. `crud.py` now only does SQL.

services/apps/git_integration/src/crowdgit/database/crud.py
skwowet: Done — `get_repo_affiliation_registry` now returns `RepoAffiliationRegistry.from_db(...)` (or `None`).

services/apps/git_integration/src/crowdgit/database/crud.py
skwowet: Done — `upsert_repo_affiliation_registry` now takes a `RepoAffiliationRegistry` model.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,911 @@
...
+        error_message = None
+        ai_cost = 0.0
+        latest_file_path: str | None = None
+        latest_file_hash: str | None = None
+        registry = await get_repo_affiliation_registry(repository.id)
+
+        try:
+            has_interval_elapsed, _ = await self.check_if_interval_elapsed(registry)
+            if not has_interval_elapsed:
+                raise AffiliationIntervalNotElapsedError()
```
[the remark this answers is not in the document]
skwowet: `remaining_hours` is now included in the `AffiliationIntervalNotElapsedError` message (same pattern as maintainers), so it shows up in `serviceExecutions` without extra info logs.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,911 @@
...
+            self.logger.error(error_message)
+
+        finally:
+            end_time = time_module.time()
+            execution_time = Decimal(str(round(end_time - start_time, 2)))
+
+            service_execution = ServiceExecution(
+                repo_id=repository.id,
+                operation_type=OperationType.AFFILIATION,
+                status=execution_status,
```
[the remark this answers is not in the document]
skwowet: Agreed — code uses `OperationType.AFFILIATION` to match `MAINTAINER`. Updated the PR description accordingly.

services/apps/git_integration/src/crowdgit/database/crud.py
skwowet: Changed to `-> None`. With `ON CONFLICT DO NOTHING`, returning `len(rows)` was the attempted count, not inserted. If we need an insert count in metrics later we can add `RETURNING`.

services/apps/git_integration/src/crowdgit/database/crud.py
skwowet: Same fix — changed to `-> None` for the same reason as above.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: I think we can use it in the interval error message (maintainer parity) rather than adding a separate log line.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: Removed the `= ""` default — callers always pass `repository.url`. Also removed unused `repo_url` from `parse_affiliations` / `resolve_snapshot` since it was not used in the parse path.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: Same — removed the default here as well.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: Removed the `CommandExecutionError` handler. Failures now fall through to the generic `Exception` handler and get logged. Kept the separate `FileNotFoundError` case for the rg-not-installed message.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: Removed the sort here — `find_known_file_matches` already returns sorted results.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: Done

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
skwowet: Good catch on both points.

1. **Single-match check** — fixed. We now run `is_text_file_path` before trusting a lone match, so a same-named binary file (e.g. `sigs.bin`) won't be auto-accepted:

```313:317:services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
        if len(matches) == 1:
            only_match = matches[0]
            if self.is_text_file_path(only_match):
                self.logger.info(f"Affiliation file: {only_match}")
                return only_match, ai_cost
```

2. **Consolidated rg filters** — updated `find_files_by_known_name` to build extension-specific `--iglob` patterns from `TEXT_FILE_EXTENSIONS` (mirroring maintainer's approach) instead of the broad `**/{name}.*`, then post-filter with `path_matches_known_name` + `is_text_file_path`.

One intentional difference from maintainers: we kept the known-name search recursive (`**/`) rather than root-only, since affiliation files (e.g. `sigs.yaml`) often live in subdirectories, not just repo root. Root-only listing is still used separately as the AI-picker fallback (`list_root_text_files`).

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +525,260 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+        SELECT "filePath", "fileHash", "status", "snapshot", "lastRunAt"
+        FROM git."repoAffiliationRegistry"
+        WHERE "repoId" = $1
+    """
+    result = await fetchrow(sql_query, (repo_id,))
+    if not result:
+        return None
+
+    row = dict(result)
+    row["repoId"] = repo_id
```
mbani01: We can get repoId from ` git."repoAffiliationRegistry"`

```suggestion
    sql_query = """
        SELECT "repoId", "filePath", "fileHash", "status", "snapshot", "lastRunAt"
        FROM git."repoAffiliationRegistry"
        WHERE "repoId" = $1
    """
    result = await fetchrow(sql_query, (repo_id,))
    if not result:
        return None

    row = dict(result)
```

mbani01 (dismissed):

services/apps/git_integration/src/crowdgit/database/crud.py
skwowet: Done — `get_repo_affiliation_registry` now selects `"repoId"` from `git."repoAffiliationRegistry"` and passes the row straight to `RepoAffiliationRegistry.from_db()`, so we no longer inject `repoId` manually.

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +525,258 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+    """
+    result = await fetchrow(sql_query, (repo_id,))
+    if not result:
+        return None
+
+    return RepoAffiliationRegistry.from_db(dict(result))
+
+
+async def upsert_repo_affiliation_registry(registry: RepoAffiliationRegistry) -> None:
+    snapshot_json = registry.snapshot_for_db()
```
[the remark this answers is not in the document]
skwowet: Not applicable — asyncpg handles jsonb read/write with Python list/dict natively. Snapshot caching and upsert are consistent with that. The commit_service orjson.loads example is a different code path (in-memory activity tuples, not jsonb reads from this table).

services/apps/git_integration/src/crowdgit/models/affiliation_info.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @staticmethod
+    def _parse_snapshot(snapshot) -> list[AffiliationInfoItem] | None:
+        if isinstance(snapshot, dict) and "affiliations" in snapshot:
+            snapshot = snapshot["affiliations"]
+        try:
+            return _SNAPSHOT_ADAPTER.validate_python(snapshot)
+        except ValidationError as error:
+            logger.warning(f"Invalid affiliation snapshot in registry, will re-parse: {error}")
+            return None
```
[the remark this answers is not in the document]
skwowet: Not applicable — asyncpg handles jsonb read/write with Python list/dict natively. Snapshot caching and upsert are consistent with that. The commit_service orjson.loads example is a different code path (in-memory activity tuples, not jsonb reads from this table).

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,916 @@
...
+            output = await run_shell_command(
+                ["rg", "--files", "--hidden", *glob_args, "."],
+                cwd=repo_path,
+            )
+        except FileNotFoundError:
+            self.logger.warning("Ripgrep not found, known filename search is unavailable")
+            return []
+        except Exception as e:
+            self.logger.warning(f"Known filename search failed for {known_name!r}: {repr(e)}")
+            return []
```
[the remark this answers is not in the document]
skwowet: Acknowledged as log-noise only. rg exit 1 means "no matches"; we return [] correctly. Can align with maintainer's CommandExecutionError handler to avoid ERROR-level logs on the expected path — not a functional issue.

services/apps/git_integration/src/crowdgit/database/crud.py
```diff
@@ -524,3 +525,258 @@ async def save_service_execution(service_execution: ServiceExecution) -> None:
...
+            "dateStart",
+            "dateEnd",
+            "title",
+            source,
+            verified,
+            "createdAt",
+            "updatedAt"
+        )
+        VALUES ($1, $2, NULL, NULL, NULL, $3, false, NOW(), NOW())
+        ON CONFLICT ("memberId", "organizationId", "dateStart", "dateEnd") DO NOTHING
```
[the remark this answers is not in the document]
skwowet: Valid for when MO writes are enabled. Soak keeps inserts commented out. We'll align ON CONFLICT with the partial unique index (ix_unique_member_org_no_date_nulls) before turning writes on — same pattern as DAL.

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,945 @@
...
+
+            self.logger.info(f"Finished affiliations from {latest_file_path}")
+
+        except AffiliationIntervalNotElapsedError as e:
+            self.logger.info(e.error_message)
+
+        except AffiliationFileNotFoundError as e:
+            ai_cost = e.ai_cost
+            self.logger.info(e.error_message)
+
```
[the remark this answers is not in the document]
skwowet: This is intentional to avoid error noise, as ‘not found’ is common

services/apps/git_integration/src/crowdgit/models/affiliation_info.py
```diff
@@ -0,0 +1,118 @@
...
+
+        snapshot = row.get("snapshot")
+        if snapshot is not None:
+            row["snapshot"] = cls._parse_snapshot(snapshot)
+
+        return cls(**row)
+
+    @staticmethod
+    def _parse_snapshot(snapshot) -> list[AffiliationContributorEntry] | None:
+        if isinstance(snapshot, str | bytes):
```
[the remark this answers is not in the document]
skwowet: N/A — Python 3.10+ supports isinstance with union types. Prod snapshots parse correctly

services/apps/git_integration/src/crowdgit/services/affiliation/affiliation_service.py
```diff
@@ -0,0 +1,979 @@
...
+            organization = row.organization
+            is_unaffiliated = organization.is_unaffiliated
+            domain = cls._strip(organization.domain)
+
+            if is_unaffiliated:
+                stint = AffiliationOrganizationStint(
+                    name="Individual",
+                    domain="nonameaccount.com",
+                    date_start=cls._parse_optional_date(organization.date_start),
```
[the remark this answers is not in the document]
skwowet: Intentional — prod placeholder org is `nonameaccount.com`. `individual-noaccount.com` is not in CDP; using it would break org lookup on apply

PR #4305, opened 2026-07-03
fix: decimal overflow in enriched packages copy

epipav:
> [!NOTE]
> **Low Risk**
> Narrow change to one analytics pipe node; behavior is equivalent for valid dates and only excludes corrupt or extreme timestamps.
> 
> **Overview**
> Fixes the **ossPackages_enriched_release** Tinybird pipe so the scheduled COPY into `ossPackages_enriched_ds` no longer fails on decimal overflow.
> 
> **secondPublished** no longer uses `arraySort` with negated `toUnixTimestamp(publishedAt)`; it now takes the second-most-recent date via `arrayReverseSort(groupArray(publishedAt))`, avoiding numeric overflow from timestamp negation.
> 
> The **versions** filter now drops out-of-range `publishedAt` values (before epoch and more than one day in the future) in addition to nulls, so bad dates do not propagate into **releaseCadence** scoring downstream.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 726fd36f59ccdcee0493dde4911a70cb020e7016. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joanagmaia:
@epipav if this is already in production can we merge it? Some Tinybird PRs are becoming stale, and might lead to confusion on what's in production or not. Thanks!

PR #4405, opened 2026-07-27
perf: precompute collection_contributor_dependency filtered variants (IN-1196)

gaspergrom:
## Summary
- Extends the collection-page precompute work to cover the contributor-dependency widget's real filtered traffic — production telemetry showed 86% of requests carry a date range, and 98% of those match one of the 8 fixed date-picker presets (not the "no filters at all" case the original v1 precompute covered).
- Adds a 16-variant precomputed matrix (8 date presets × 2 `includeCollaborations` states) for `collection_contributor_dependency`, replacing the old v1 single-variant (unfiltered-only) precomputed path.
- Splits `collection_contributor_dependency_copy` and `collection_contributors_leaderboard_copy` each into 8 `presetKey`-scoped copy pipes (`past90days`, `past180days`, `past365days`, `previousQuarter`, `previousYear`, `previous5years`, `previous10years`, `allTime`), mirroring the existing `collection_insights_aggregate_copy` pattern.
- Rewrites `collection_contributor_dependency_copy_ds` and `collection_contributors_leaderboard_copy_ds` as append + `snapshotId` datasources so a daily run doesn't clobber the prior day's rows mid-query.
- Updates the serving pipes (`collection_contributor_dependency.pipe`, `collection_contributors_leaderboard.pipe`) to route between the precomputed path (requires `presetKey`, no `repos`/`platform`/`activity_type`, `limit`=100) and the live `GROUP BY` path at template-render time — ClickHouse evaluates all `UNION ALL` branches regardless of `WHERE` guards, so routing has to happen at render time, not query time.
- Deletes the superseded v1 `collection_contributor_dependency_copy.pipe` / `collection_contributor_dependency_copy_ds.datasource`.

## Why
At collection scale (e.g. `cncf` has 149,677 distinct contributors), the live `GROUP BY af.memberId` path for the contributor-dependency widget was triggering Tinybird production rate-limit/concurrency alerts under load. The original precompute only covered fully-unfiltered requests, but real traffic is overwhelmingly filtered by one of the date picker's 8 fixed presets — so the fast path almost never activated. This closes that gap.

## Changes
- `services/libs/tinybird/pipes/collection_contributor_dependency.pipe` — adds precomputed-path routing node, updated DESCRIPTION.
- `services/libs/tinybird/pipes/collection_contributors_leaderboard.pipe` — same routing pattern (was already partially split in an earlier phase; this makes the presetKey contract consistent with contributor-dependency).
- `services/libs/tinybird/pipes/collection_insights_aggregate.pipe` — minor consistency update.
- 16 new `*_copy_<presetKey>.pipe` files (8 for contributor-dependency, 8 for leaderboard).
- `collection_insights_aggregate_copy.pipe` — new copy pipe for that datasource.
- 3 rewritten/new `*_copy_ds.datasource` files (append + snapshotId schema).
- Deletes `collection_contributor_dependency_copy.pipe` (v1, superseded).

## Deploy order
Must deploy before the companion insights PR (#2041) — that PR sends `presetKey` values that need to already resolve against a live pipe here.

## Test plan
- [x] Validated pipe/datasource SQL and schema changes via Tinybird MCP against production data
- [x] Deployed through staging then production via `tb` CLI
- [x] Verified all 16 contributor-dependency variants and 16 leaderboard variants present with matching row counts for a real collection (`cncf`)
- [x] Root-caused and fixed a `past365days` zero-rows gap post-deploy (one pipe instance's daily cron hadn't fired yet after the redeploy reset its run history) — manually triggered, verified correct data end-to-end

JIRA: IN-1196

### DIFF
--- a/services/libs/tinybird/datasources/collection_contributor_dependency_copy_ds.datasource
+++ b/services/libs/tinybird/datasources/collection_contributor_dependency_copy_ds.datasource
@@ -1,0 +1,66 @@
+DESCRIPTION >
+    - `collection_contributor_dependency_copy_ds` is the Phase 3 replacement for a retired v1
+    datasource of the same name that covered only a single unfiltered variant: it now holds
+    precomputed bus-factor (contributor dependency) results for every collection across the same
+    fixed 16-variant filter matrix as `collection_contributors_leaderboard_copy_ds` (8 date-range
+    presets x 2 `includeCollaborations` states).
+    - Created via `collection_contributor_dependency_copy.pipe` (`TYPE COPY`, daily schedule,
+    `COPY_MODE append`). Unlike the retired v1 pipe (and unlike the 8
+    `collection_contributors_leaderboard_copy_<presetKey>.pipe` pipes), this copy pipe does NOT scan
+    `activityRelations_deduplicated_cleaned_bucket_union` itself - the bus-factor computation
+    (running-total percentage + 51% cutoff + rank) is a pure downstream transform of
+    `collection_contributors_leaderboard_copy_ds`'s already-ranked rows, so this pipe reads that
+    table instead of repeating an expensive activity scan. This also means it carries none of the
+    fan-out-before-aggregate risk that both the retired v1 pipe and the original
+    `collection_insights_aggregate_copy.pipe` incident had - there is no raw-activity join here at
+    all.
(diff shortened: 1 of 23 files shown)


gaspergrom:
Closing without merging — abandoning this optimization work per decision to revert all Tinybird performance-optimization changes (Phases 1-3) and start fresh.

PR #4429, opened 2026-08-03
fix: remove USING joins from health score copy tree

epipav:
## Problem

The nightly `health_score_copy` COPY pipe (00:50 UTC) fails with:

```
BAD_ARGUMENTS: JOIN LEFT JOIN ... USING (segmentId) USING clause expected identifier. Actual segmentId.
```

ClickHouse 25.3+ is stricter about `JOIN ... USING` when the joined sides are subqueries with aliased columns — which every `health_score_*` metric pipe is (aggregated `GROUP BY` subqueries). Since Tinybird inlines all referenced pipes into one query at copy time, the whole query tree is affected, not just the top-level pipe.

## Change

Convert every `USING (key)` join in the `health_score_copy` query tree to an explicit `ON left.key = right.key`, preserving join type and keys:

- `health_score_copy.pipe` — 13 joins in `health_score_copy_data` (11× `segmentId`, 2× `id`)
- `health_score_retention.pipe` — 1 join (`ON cur.segmentId = prev.segmentId`)
- `health_score_security.pipe` — 1 inner join on `repo`; the select-list `repo` is qualified as `security_deduplicated_merged_ds.repo AS repo` since both sides keep the column after an `ON` join

Output columns (names, order), join semantics, and the `health_score_copy_ds` target schema are unchanged.

## Validation

- On tinybird-local (ClickHouse 25.8.23.6) with the full dependency closure pushed and `insightsProjects` fixtures: the rewritten 13-join node returns **byte-identical results** to the original (38 rows, matching hash), and the fixed retention/security pipes push and execute cleanly.
- The production `BAD_ARGUMENTS` originates in Tinybird cloud's copy-job execution path and does not reproduce through the local classic image; the rewrite follows Tinybird's recommended remediation.
- `tb fmt --diff` clean on all three files (tinybird-cli 5.x, matching CI).

## Deferred (follow-ups)

9 other pipes still contain the same `USING` construct and will likely need the same conversion as they hit the stricter path: `repo_health_score_copy`, `repo_health_score_retention`, `project_insights_copy`, `insights_projects_populated_copy`, `org_dash_metric_copy_pipe`, `collections_oss_index`, `categories_oss_index`, `category_groups_oss_index`, `activity_heatmap_by_weekday_and_2hours_blocks`. Deliberately excluded here to keep this fix minimal.


epipav:
@cursor review

gaspergrom (approved):

PR #4464, opened 2026-08-12
fix: round geo distribution percentage to 2 decimal places

emlimlf:
## Summary
- `contributors_geo_distribution.pipe` and `organizations_geo_distribution.pipe` now round the percentage field to 2 decimal places instead of the nearest integer
- Fixes countries under 1% share showing as `0%` in the geographical distribution widget, now that the widget lists all countries (IN-1225)

## Test plan
- [ ] Deploy pipe to Tinybird and confirm small-share countries return non-zero fractional percentages
- [ ] Confirm percentages still sum to ~100


joanagmaia (approved):
LGTM

PR #4476, opened 2026-08-13
feat: filter deleted member enrchiment worker (CM-1367)

ulemons:
## Summary

Refactors `members_enrichment_worker`'s org-affiliation reconciliation from delete-and-recreate to update-in-place. Today, every enrichment run deletes every non-UI/non-project-registry `memberOrganizations` row and recreates whatever the provider supplies, which causes two bugs: (1) an org a human manually deleted because it was wrong (e.g. the "Dusky'z" case — OpenDaylight/Apache/Karaf contributions wrongly attributed to a non-existent company instead of PANTHEON.tech) gets resupplied and recreated by the provider on the next run, and (2) `verified`/`verifiedBy` status is lost every time, since the recreated row is always unverified. This PR makes enrichment update rows in place, so verification and human deletes survive reruns.

## Changes

- New `"deletedBy"` column on `"memberOrganizations"` (migration, additive, no backfill). Set only on human-initiated deletes; `NULL` means the delete was enrichment's own rebuild and stays recoverable (protects against provider-glitch data loss).
- `deleteMemberOrganizations()` (`data-access-layer`) takes an optional `deletedBy`, wired through the human-delete paths: `verifyMemberWorkExperience.ts` (unverify) and `memberOrganizationsService.ts#delete` (CDP UI org delete). The enrichment worker's own delete path (`deleteMemberOrgById`) is untouched — it must keep leaving `deletedBy` `NULL`.
- `fetchMemberDataForLLMSquashing`'s tombstone CTE now keys off `"deletedAt" IS NOT NULL AND "deletedBy" IS NOT NULL` instead of guessing from `source`; also selects `verified`/`"verifiedBy"` so the worker can protect verified rows.
- Extracted `workExperienceReconciliation.ts` from `enrichment.ts` (pure, unit-testable). `prepareWorkExperiences` now reconciles old vs. new orgs by `(organizationId, normalized title)`: matched rows are updated in place (only the fields that changed), unmatched old rows are soft-deleted (`deletedBy` `NULL`, recoverable), unmatched new entries are inserted, and orgs with `deletedBy` set are never recreated. UI, project-registry, and `verified = true` rows are never touched.
- `affiliationNeedsRefresh` narrowed to fire only on an actual date change, not a cosmetic title-only update — cuts down noise on `activityRelations` refreshes.
- `updateMemberOrg` now sets `"updatedAt" = now()` on every update.
- Added unit tests for `prepareWorkExperiences` / `hasMemberOrganizationTimelineChange` (14 cases): verified-row protection, tombstone respect, provider-glitch recoverability, no writes on unchanged payload, UI-row conflict handling in both the normal and high-confidence branches.

**Verified end-to-end locally**, not just unit tests: seeded a member with a verified affiliation, a manually-tombstoned org, and a provider-soft-deleted (non-tombstoned) org, then ran the real `processMemberSources` Temporal workflow against it. Result matched all three expectations — verified row untouched, tombstoned org never recreated, provider-glitch org recreated as a new active row.

**Known gap, intentionally out of scope — tracked in a separate ticket:** the public API v1 `DELETE /members/:memberId/work-experiences/:workExperienceId` has no actor field (it's an M2M token, so `req.actor` carries the calling LFXOne service's client id, not the acting human). Deletes through that endpoint won't be tombstoned yet. Confirmed with Nathan and Joana to handle this separately since it needs LFXOne coordination on the payload contract.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Chore / dependency update
- [ ] Documentation

## JIRA ticket
[CM-1367](https://linuxfoundation.atlassian.net/browse/CM-1367)

[CM-1367]: https://linuxfoundation.atlassian.net/browse/CM-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


services/apps/members_enrichment_worker/vitest.config.ts
skwowet, line 1: Root `pnpm test:server` already picks up `services/**/*.test.ts`, so this local Vitest setup is redundant. Please drop it and let CI run the tests from the repo root.

services/apps/members_enrichment_worker/package.json
```diff
@@ -35,6 +36,7 @@
   "devDependencies": {
     "@types/node": "^20.8.2",
     "@types/uuid": "~9.0.6",
-    "nodemon": "^3.0.1"
+    "nodemon": "^3.0.1",
+    "vitest": "^3.2.4"
```
skwowet: Can we drop vitest here as well?

services/apps/members_enrichment_worker/package.json
```diff
@@ -10,7 +10,8 @@
     "lint": "npx eslint --ext .ts src --max-warnings=0",
     "format": "npx prettier --write \"src/**/*.ts\"",
     "format-check": "npx prettier --check .",
-    "tsc-check": "tsc --noEmit"
+    "tsc-check": "tsc --noEmit",
+    "test": "vitest run"
```
skwowet: Same here

services/apps/members_enrichment_worker/src/activities/workExperienceReconciliation.ts
```diff
@@ -0,0 +1,289 @@
...
+  const verifiedRows = oldVersion.filter((c) => c.verified === true)
+  newVersion = newVersion.filter(
+    (e) =>
+      !verifiedRows.some(
+        (v) =>
+          normalizeTitle(v.jobTitle) === normalizeTitle(e.title) &&
+          e.identities &&
+          e.identities.some((i) => i.organizationId === v.orgId),
+      ),
+  )
```
skwowet, line 203: Verified rows should win: if the provider resends the same job with conflicting dates, drop the provider row. I don’t think this check does that on real data.

Walkthrough:

1. DB already has verified Engineer at Acme (`orgId = acme-uuid`), 2020–now.
2. Provider sends Engineer at Acme again, dates 2021–2022. Identities are only a LinkedIn handle — they never include `acme-uuid`.
3. Enrichment recognizes Acme and sets `e.organizationId = acme-uuid` on the job. It does not put that id on `e.identities[]`.
4. This filter looks for `acme-uuid` on identities, finds nothing, and inserts a second unverified row.

We still have the verified row, plus a duplicate with the provider’s dates.

`newEntry()` in the tests always puts our org id on the identity, so this path never runs.

Can we match `e.organizationId === v.orgId` (and title) instead, including the UI filters below that use the same identity check? Also, let's update the tests accordingly.

skwowet (changes requested):
Looks good overall 👌🏻 

The update-in-place and the deletedBy tombstone are the right direction. 

One question on verified-row matching (identities vs organizationId) before merge; also a small note on the extra Vitest config.

services/apps/members_enrichment_worker/src/activities/workExperienceReconciliation.ts
ulemons: Good catch, you are right! the filter was matching against `e.identities`, but enrichment resolves the org onto `e.organizationId` without ever backfilling identities[] with it, so on real data the check never fired and a verified row's conflicting resend would slip through as a duplicate.

Fixed by matching on e.organizationId === v.orgId (+ title) as you are suggesting, instead of scanning identities, and applied the same fix to the UI-row filter in the high-confidence branch and the manual-match branch below 

Also fixed `newEntry()` in the test fixture: it was putting our org id on the identity by default, which is why this path never got exercised in tests. It now only sets organizationId, matching the real shape enrichment produces.

skwowet (approved):

PR #4483, opened 2026-08-17
feat: route collection-scoped activity queries to single bucket (IN-1231)

gaspergrom:
## Summary

- Adds `activityRelations_collection_bucket_routing.pipe`, a single-bucket routing pipe that mirrors the existing 10-way-union pattern but for collection-scoped queries. Routes to the correct pre-resolved bucket (0-9) based on the optional `bucketId` parameter.
- Updates `activities_filtered.pipe` to accept optional `bucketId` alongside `collectionSlug`. When both are provided, routes directly to the single-bucket pipe instead of the union, improving query performance for pre-resolved collections. Fully backward compatible — callers passing only `collectionSlug` continue to use the existing union pipe.
- This PR is inert without the companion insights PR (`feat/IN-1231`) which wires up the caller-side logic to pass `bucketId`. Can merge independently or in any order relative to the insights PR.

Pipes already validated and deployed to Tinybird production by `crowd-tinybird-manager`:
- Bucket-0 routing tested against real collection `cloud-management-platforms` (34.1M rows); correct row counts confirmed
- All 20+ downstream consumer pipes (health metrics, leaderboards, collections API) verified intact
- Backward compatibility confirmed for existing `collectionSlug`-only callers

## JIRA

IN-1231 — Route collection-scoped activity queries to single bucket (insights-driven)

### DIFF
--- a/services/libs/tinybird/pipes/activities_filtered.pipe
+++ b/services/libs/tinybird/pipes/activities_filtered.pipe
@@ -37,9 +37,7 @@ SQL >
     where
         {% if defined(collectionSlug) %}
             segmentId IN (SELECT segmentId FROM segments_filtered_by_collection)
-            {% if defined(bucketId) %}
-                AND a.collectionSlug = {{ String(collectionSlug) }}
-            {% end %}
+            {% if defined(bucketId) %} AND a.collectionSlug = {{ String(collectionSlug) }} {% end %}
         {% else %} segmentId = (SELECT segmentId FROM segments_filtered)
         {% end %}
         {% if defined(startDate) %}
--- a/services/libs/tinybird/pipes/activities_filtered.pipe
+++ b/services/libs/tinybird/pipes/activities_filtered.pipe
@@ -16,8 +16,9 @@ DESCRIPTION >
     - `activity_types`: Optional array of activity types (e.g., ['authored-commit', 'co-authored-commit'])
     - `includeCodeContributions`: Optional boolean to include code contribution activities. Defaults to 1. Set to 0 to exclude. Inherited from activityTypes_filtered.
     - `includeCollaborations`: Optional boolean to include or exclude collaboration activities. Inherited from activityTypes_filtered.
-    - `bucketId`: Optional pre-resolved collection bucket id (from `collection_buckets.pipe`). When given alongside
-    `collectionSlug`, routes directly to the single matching bucket table instead of scanning the 10-way union.
+    - `bucketId`: Pre-resolved collection bucket id (from `collection_buckets.pipe`), required alongside `collectionSlug`.
+    Routes directly to the single matching bucket table. If `collectionSlug` is given without `bucketId`, the pipe
+    returns no rows rather than falling back to a 10-way union scan — callers must resolve `bucketId` first.
     - Response: `id` (activityId), `timestamp`, `type`, `platform`, `memberId`, `organizationId`, `segmentId`.
     - This pipe is consumed by many of downstream pipes and widgets across the platform for consistent activity filtering.
     - Performance is optimized through proper sorting keys on `segmentId`, `timestamp`, `type`, `platform`, and `memberId` in the source datasource.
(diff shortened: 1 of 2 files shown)


services/libs/tinybird/pipes/activities_filtered.pipe
```diff
@@ -27,12 +29,17 @@ SQL >
...
+            activityRelations_collection_bucket_routing
+        {% elif defined(collectionSlug) %} activityRelations_collection_bucket_union
         {% else %} activityRelations_bucket_routing
         {% end %} as a
     where
         {% if defined(collectionSlug) %}
             segmentId IN (SELECT segmentId FROM segments_filtered_by_collection)
+            {% if defined(bucketId) %}
+                AND a.collectionSlug = {{ String(collectionSlug) }}
+            {% end %}
```
epipav, line 42: I don't think activities have a `collectionSlug` field. Do we need this filter at all?

services/libs/tinybird/pipes/activities_filtered.pipe
```diff
@@ -27,12 +29,17 @@ SQL >
     %
     SELECT activityId as id, timestamp, type, platform, memberId, organizationId, segmentId
     FROM
-        {% if defined(collectionSlug) %} activityRelations_collection_bucket_union
+        {% if defined(collectionSlug) and defined(bucketId) %}
+            activityRelations_collection_bucket_routing
+        {% elif defined(collectionSlug) %} activityRelations_collection_bucket_union
```
epipav: Do we need to support this? Pipes that need a bucket union should directly ask for it, imo, for this particular pipe, we can always use the routings

epipav (commented):
added two comments

services/libs/tinybird/pipes/activityRelations_collection_bucket_routing.pipe
```diff
@@ -0,0 +1,18 @@
...
+        {% elif bucketId == '1' %} activityRelations_collection_deduplicated_cleaned_bucket_1_ds
+        {% elif bucketId == '2' %} activityRelations_collection_deduplicated_cleaned_bucket_2_ds
+        {% elif bucketId == '3' %} activityRelations_collection_deduplicated_cleaned_bucket_3_ds
+        {% elif bucketId == '4' %} activityRelations_collection_deduplicated_cleaned_bucket_4_ds
+        {% elif bucketId == '5' %} activityRelations_collection_deduplicated_cleaned_bucket_5_ds
+        {% elif bucketId == '6' %} activityRelations_collection_deduplicated_cleaned_bucket_6_ds
+        {% elif bucketId == '7' %} activityRelations_collection_deduplicated_cleaned_bucket_7_ds
+        {% elif bucketId == '8' %} activityRelations_collection_deduplicated_cleaned_bucket_8_ds
+        {% elif bucketId == '9' %} activityRelations_collection_deduplicated_cleaned_bucket_9_ds
+        -- fallback, should never happen
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 7db18a273: removed non-explanatory comment per code quality guidelines (comments should only explain non-obvious invariants, not restate what the code does).

services/libs/tinybird/pipes/activities_filtered.pipe
gaspergrom: @epipav The collection-side activity tables (activityRelations_collection_deduplicated_cleaned_bucket_*) do have a collectionSlug column — it's the first sort key in their ENGINE_SORTING_KEY. This filter is a correctness check applied only when bucketId is pre-resolved (line 40 condition). The bucket_union pipe already filters by collectionSlug internally, so this line is skipped in that case (bucketId undefined). When bucketId IS defined, we route to a single bucket table and this filter ensures data consistency.

services/libs/tinybird/pipes/activities_filtered.pipe
gaspergrom: @epipav The bucket union path exists to avoid requiring callers to pre-resolve bucketId. This is a convenience tradeoff: callers can either (1) pre-resolve bucketId and route directly to one bucket table (fast path), or (2) omit bucketId and let the union handle it transparently (simpler API, leverages ClickHouse's sort-key pruning). Requiring all callers to fetch and provide bucketId would add friction. If this flexibility isn't worth the extra code path, we could simplify to require bucketId always — happy to discuss the tradeoff.

gaspergrom: Fixed in 0b6e3f090: removed the union fallback. When `collectionSlug` is given without `bucketId`, the pipe now returns no rows instead of scanning the 10-way union — bucketId is effectively required for collection queries.

epipav (approved):

PR #4502, opened 2026-08-24
fix: tombston in inferMemberOrganizationStintChanges (CM-1367)

ulemons:
## Summary

Fixes DE-1016/DE-1021: a human-deleted `email-domain` sourced org affiliation can reappear on a member profile once new matching activity comes in, because the `infer-member-organization-stint-changes` cron job only checked `deletedAt` and never `deletedBy`. PR #4476 (CM-1367) already fixed this for provider/enrichment-sourced affiliations, but the email-domain stint-inference path is a separate codepath that PR didn't touch. This PR closes that gap by making a human tombstone (`deletedBy` set) permanently block recreation of that org affiliation, regardless of whether the new activity date falls inside the deleted row's date range.

## Changes

- Added `deletedBy?: string` to `IMemberOrganization` (`services/libs/types/src/organizations.ts`) so the field can flow through to the stint-inference logic.
- Added `"deletedBy"` to the SELECT column list in `fetchMemberOrganizationsBySource` (`services/libs/data-access-layer/src/members/organizations.ts`) — the cron job already fetches deleted rows via `withDeleted: true`, it just wasn't reading who deleted them.
- `inferMemberOrganizationStintChanges` (`services/libs/common_services/src/services/member-organization.ts`) now computes a `tombstonedOrgIds` set from rows where both `deletedAt` and `deletedBy` are set, and skips any stint insert/extend for those `organizationId`s outright. This is intentionally stronger than the existing `deletedRows` suppression, which only blocks recreation for activity dates that intersect the deleted row's own date range — a human delete should hold regardless of date, while a system/provider delete (no `deletedBy`) keeps the narrower, date-scoped suppression.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Chore / dependency update
- [ ] Documentation

## JIRA ticket
[CM-13267](https://linuxfoundation.atlassian.net/browse/CM-1367)

### DIFF
--- a/services/apps/cron_service/src/jobs/inferMemberOrganizationStintChanges.job.ts
+++ b/services/apps/cron_service/src/jobs/inferMemberOrganizationStintChanges.job.ts
@@ -119,7 +119,6 @@ const job: IJobDefinition = {
 
         processed++
       } catch (err) {
-        // Rare race past dropStaleOrganizationDates(): leave the entry queued, next tick self-heals.
         if ((err as { code?: string })?.code === '23503') {
           ctx.log.warn(err, { memberId }, 'Stint change referenced missing organization.')
           continue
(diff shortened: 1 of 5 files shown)


skwowet (dismissed):
LGTM

PR #4608, opened 2026-09-11
chore: pin node version via asdf .tool-versions (CM-1444)

themarolt:
## Summary
Adds a project-local `.tool-versions` so asdf resolves the repo's required node version automatically.

## Changes
- Add `.tool-versions` (`nodejs 24.20.0`) — asdf reads `.tool-versions`, not `.nvmrc`, so without this file a global asdf node pin silently wins inside the project instead of the repo's required version (24, per `.nvmrc` / `engines.node`).
- Kept `.nvmrc` and `engines.node` as-is (major-only / range) — this file is the exact-pin equivalent for asdf/mise users specifically, not a change to the repo's version policy.

## Type of change
- [x] Chore / dependency update

## JIRA ticket
https://linuxfoundation.atlassian.net/browse/CM-1444


gaspergrom (approved):
Confirmed nodejs 24.20.0 satisfies engines.node (>=24.0.0) and matches .nvmrc's major version 24, and there's no other .tool-versions or .nvmrc elsewhere in the repo this could conflict with. LGTM.

PR #4610, opened 2026-09-11
docs: add fix-vulns-git-integration skill

mbani01:
## Summary

Adds an agentic vulnerability-triage skill scoped to the `git_integration` worker (Python deps via uv + the two vendored Go modules), invocable as `/fix-vulns-git-integration`.

What it does per run:
- Fetches open Dependabot alerts for the three git_integration manifests, dedupes by package (case-insensitive), drops stale alerts already fixed in main, and correlates with open Dependabot PRs.
- Classifies each bump `safe` vs `needs-human` via parallel break-risk agents (usage scan + changelog evidence; uncertainty is always needs-human, majors never auto).
- Validates safe candidates locally on a combined branch merged with latest main: ruff + pytest, plus a full git-integration image build (compiles both Go binaries, `uv sync --frozen`) — repo CI doesn't cover this worker, so this is the real gate.
- Updates validated PR branches server-side ("Update branch") so they're one click from merge, posts a "safe to merge" review summary to Slack (human merges — the skill never merges, never dismisses alerts), and opens a batch PR for alerts Dependabot has no PR for.
- Deploy dispatch only on explicit user confirmation, staging before production.

Also registers the skill in `.claude/rules/skill-guidance.md`.

## Test plan

First e2e run on 2026-09-10 against the 20 open critical alerts: 3 PRs validated + branch-updated (#4587 gitpython — since merged, #4318 x/crypto, #4552 grpc), 3 packages correctly routed to needs-human (pgx batch-decode churn, software_value Go 1.24→1.25 directive), summary posted to #cdp-ins-github-pr-queue.


joanagmaia (approved):
LGTM

PR #4656, opened 2026-09-16
fix: return partial data on github graphql errors instead of failing sync (CM-1445)

mbani01:
## Summary
- `githubGraphql` no longer throws on any non-empty GraphQL `errors` array. It now logs the errors (as `warn` when the error `type` is `FORBIDDEN`, `error` otherwise) and returns the partial `data` GitHub already sent, matching the pattern the Nango-based GitHub integration already uses in production.
- Root cause: the `USER_FIELDS` fragment requests `organizations(first: 5)` on every `User` it touches (fork owners, PR authors, PR timeline actors). When that user belongs to an org with GitHub's IP allow-list feature enabled, GitHub returns a `FORBIDDEN` error for that nested field — unrelated to the repo actually being synced (e.g. `kubernetes/kubernetes` forks/PR syncs failing because of `quintoandar`/`Skyscanner`, third-party orgs some contributor belongs to). Per GraphQL semantics this only nulls that one field; the rest of the response (fork/PR/user core data) is still populated, so failing the whole sync unit was discarding data GitHub had already returned.
- Also removes the `stars` sync (`syncs/stars.ts`, `graphql/stars.ts`, `mappers/star.ts`) and its commented-out registration in `index.ts` — it was already disabled/unregistered and unreferenced anywhere else.

## Test plan
- [x] `npx tsc --noEmit` passes in `services/libs/connectors`
- [x] `npx eslint --ext .ts src --max-warnings=0` passes in `services/libs/connectors`
- [ ] No existing test harness in this package to add a regression test to (flagged separately, left as-is per request)


mbani01:
Acknowledged (Copilot + Cursor Bugbot both flagged the same point): `githubGraphql` now forgives any GraphQL error type, not just `FORBIDDEN` — `isForbidden` only changes log severity, it doesn't gate whether the partial `data` is trusted. This is a deliberate match to the Nango GitHub integration's existing production behavior (e.g. `pull-request-comments.ts` logs-and-continues on any error type too), which was the explicit scope of this fix (CM-1445). Tracked as backlog item 22 in `connectors-poc-backlog.md` for a future pass that would only treat `FORBIDDEN` as safe-to-continue and rethrow on other error types.

PR #4658, opened 2026-09-16
build: add docs-readiness-worker build tooling (IN-1316)

gaspergrom:
## Summary

- Adds local/CI build tooling for the `docs-readiness-worker` Temporal service: a multi-stage `Dockerfile.docs_readiness_worker` (builder + slim runner, matching the existing worker Dockerfiles), a Compose service definition (`docs-readiness-worker.yaml`, with prod and `-dev` hot-reload variants on the `docs-readiness` task queue), and a build env file pointing at the OCI registry.
- Registers `docs-readiness-worker` in `scripts/cli`'s `IGNORED_SERVICES` lists for `service-restart-fe-dev` and `clean-start-fe-dev` so local frontend-dev workflows don't try to restart/rebuild it unnecessarily.
- Scope is build tooling only — no application code for the worker itself changes in this PR.

## JIRA

IN-1316 — crowd.dev: build tooling for docs-readiness-worker


scripts/services/docker/Dockerfile.docs_readiness_worker
```diff
@@ -0,0 +1,23 @@
+FROM node:24-bookworm-slim as builder
+
+RUN apt-get update && apt-get install -y python3 make g++ --no-install-recommends && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /usr/crowd/app
+RUN npm install -g corepack@latest && corepack enable pnpm && corepack prepare pnpm@9.15.0 --activate
```
themarolt: this is the old template - only `star_snapshot_worker` and `connectors_worker` still pin `pnpm@9.15.0`, the other 27 Dockerfiles do `COPY ./package.json ./pnpm-workspace.yaml ./pnpm-lock.yaml ./` + `RUN corepack enable && corepack prepare --activate`, which picks up `packageManager` (pnpm@11.22.0) from the root package.json. here the root package.json never lands in either stage, so `pnpm i --frozen-lockfile` runs in a workspace with no root manifest and the pnpm-11 lockfile gets installed by pnpm 9 - I'd expect that to break, or has it built for you locally? copying `Dockerfile.automatic_projects_discovery_worker` and swapping the app dir would also get you the `ENV pnpm_config_verify_deps_before_run=false` line on the runner (same fix applies to line 17)

scripts/services/docs-readiness-worker.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - ../../backend/.env.dist.local
+      - ../../backend/.env.dist.composed
+      - ../../backend/.env.override.local
+      - ../../backend/.env.override.composed
+    environment:
+      <<: *env-args
+    hostname: docs-readiness-worker
+    networks:
+      - crowd-bridge
+    volumes:
```
themarolt, line 45: the other `-dev` services mount every lib under `services/libs`, this one mounts three - `dev` runs nodemon watching `../../libs`, so edits to `@crowd/logging` (a direct dep) or the archetypes won't reload. intentional, or just trimmed down?

themarolt (changes requested):

scripts/services/docker/Dockerfile.docs_readiness_worker
```diff
@@ -0,0 +1,23 @@
...
+
+COPY ./services ./services
+RUN pnpm i --frozen-lockfile
+
+FROM node:24-bookworm-slim as runner
+
+WORKDIR /usr/crowd/app
+RUN npm install -g corepack@latest && corepack enable pnpm && corepack prepare pnpm@9.15.0 --activate  && apt update && apt install -y ca-certificates --no-install-recommends && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /usr/crowd/app/node_modules ./node_modules
```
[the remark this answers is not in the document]
gaspergrom: Fixed in b1f15edf7: swapped to the current worker Dockerfile template (copy `package.json` + `corepack prepare --activate`, which picks up `packageManager` pnpm@11.22.0 from root, plus `pnpm_config_verify_deps_before_run=false` on the runner).

scripts/services/docker/Dockerfile.docs_readiness_worker
gaspergrom: Good catch — this was copied from the older `star_snapshot_worker`/`connectors_worker` template. Fixed in b1f15edf7: now matches `Dockerfile.automatic_projects_discovery_worker` (copies root `package.json`, `corepack prepare --activate` to pick up `packageManager` pnpm@11.22.0, and `pnpm_config_verify_deps_before_run=false` on the runner) with the app dir swapped to `docs_readiness_worker`.

scripts/services/docs-readiness-worker.yaml
gaspergrom: Just trimmed down, not intentional — fixed in b1f15edf7: now mounts the same full lib set as the other `-dev` services (including `logging`, which is a direct dep).

scripts/services/docs-readiness-worker.yaml
```diff
@@ -0,0 +1,53 @@
...
+    networks:
+      - crowd-bridge
+
+  docs-readiness-worker-dev:
+    build:
+      context: ../../
+      dockerfile: ./scripts/services/docker/Dockerfile.docs_readiness_worker
+    command: 'pnpm run dev'
+    working_dir: /usr/crowd/app/services/apps/docs_readiness_worker
+    # user: '${USER_ID}:${GROUP_ID}'
```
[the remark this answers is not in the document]
gaspergrom: Leaving this as-is: the commented `user: '${USER_ID}:${GROUP_ID}'` line is the established convention across this compose setup (present in 33 of the 46 service yaml files, including recently-added ones), used as an opt-in override for local UID/GID mismatches. Removing it here would make this file inconsistent with the rest of the service definitions it was modeled on, and this PR's scope is matching the existing build-tooling pattern, not changing it repo-wide.

themarolt (approved):
both fixed - dockerfile matches the current template now and the -dev mounts are the full lib set. lgtm

